PR #61, opened 2019-08-21
first try at splitting id functionality

JaapWijnen:
I made a first try at splitting the id and field functionality from the database query functionality of model. Could be interesting, let me know what you think @tanner0101 

Some context: 
The idea here is to split the database related functionality of the `Model` protocol from the property wrappers around fields and the id property. I've done this by wrapping these in a protocol called `Schema` for now. This would be useful when working with third-party RESTful API's for example or in a microservices setup.


JaapWijnen:
There are still some types that should be renamed to be correct, such as _ModelCodingKey. And there are still some references to database related types which I'm not sure how to decouple. Such as `AnyProperty` and `AnyID` containing a reference to `DatabaseOutput`. However I thought this could be an interesting first try at splitting the functionality :)

PR #65, opened 2019-09-04
Conformed Parent to Filterable protocol

calebkleveter:
This allows you to query models like this:

```swift
User.query(on: database).filter(\.$agency == agencyID).all()
```


calebkleveter:
Oh, if that works then I agree, just leave it. I though that would create a query that said something like this:

```swift
SELECT * FROM users WHERE users.agency_id.id == 42
```

But if not, then we can leave it.

PR #97, opened 2019-11-30
Don't generate bad SQL for `[].create(on: db)`

gwynne:
When the CRUD method `create(on:)` is invoked on an empty `Array`, the result is usually a syntactically invalid SQL query (e.g. `INSERT INTO relation (field1, field2, field3) VALUES ` for MySQL, similar for Postgres). This PR changes the behavior so that such an invocation immediately returns success without doing anything.

Why success?

- What error would we throw? What exactly failed?
- Success seems to be the more intuitive result; consider a RESTful service which considers an empty input array to be valid. If an error was thrown for this case, all such services would have to manually check for that condition. Those who want to error should still be checking for it anyway, since user input is never to be trusted.

One counterargument is that doing it this way hides failures caused by an invalid database connection or bad model type that would otherwise be caught. However, it seems like these kinds of problems will end up being caught anyway, just at a different time.

### DIFF
--- a/Tests/FluentKitTests/XCTestManifests.swift
+++ b/Tests/FluentKitTests/XCTestManifests.swift
@@ -6,6 +6,11 @@ extension FluentKitTests {
     //   `swift test --generate-linuxmain`
     // to regenerate.
     static let __allTests__FluentKitTests = [
+        ("testCreateEmptyModelArrayDoesntQuery", testCreateEmptyModelArrayDoesntQuery),
+        ("testDecodeMissingRequired", testDecodeMissingRequired),
+        ("testDecodeWithID", testDecodeWithID),
+        ("testDecodeWithOptional", testDecodeWithOptional),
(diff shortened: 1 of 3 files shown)


Tests/FluentKitTests/XCTestManifests.swift
```diff
@@ -6,6 +6,11 @@ extension FluentKitTests {
     //   `swift test --generate-linuxmain`
     // to regenerate.
     static let __allTests__FluentKitTests = [
+        ("testCreateEmptyModelArrayDoesntQuery", testCreateEmptyModelArrayDoesntQuery),
```
twof: This is outside of the scope of this PR, but `swift test --generate-linuxmain` probably ought to be run by CI

gwynne: Maybe I just like running it myself 🙂

PR #125, opened 2019-12-29
Ensure createdAt/updatedAt timestamps equal on creation

namolnad:
I noticed that there are currently two separate calls to `touchTimestamps` for the Array `create` extension. To ensure the timestamps are completely equal at the time of creation, these should be set to the same underlying `Date` instance, which is only the case if multiple triggers are passed to a single `touchTimestamps` call. This PR combines the two calls into one call with multiple triggers.

(This also cleans up some whitespace and an unused variable, but I can leave that stuff out if it's desired)


namolnad:
@tanner0101 this is just a minor change, but wanted to see about getting your eyes on it. BTW, thanks for all your hard work on Vapor!

namolnad:
No problem, thanks @tanner0101!

PR #147, opened 2020-02-06
Convert AnyModel.schema to a class func

samalone:
This pull request changes AnyModel.schema from:

```swift
static var schema: String { get }
```

to

```swift
static func schema() -> String
```

The purpose of this change is to allow users to create generic types that inherit from Model, like:

```swift
class Annotations<ModelType: Model>: Model {
    class func schema() -> String {
        return ModelType.schema() + "_annotations"
    }
    
    @ID(key: "id")
    var id: Int?
    
    @Parent(key: "entity_id")
    var entity: ModelType

    @Field(key: "note")
    var note: String
}
```

Swift does not allow generic types to define static properties, but it does allow them to define class functions.

In addition, I have provided a default implementation of AnyModel.schema() that simply returns the name of the class. This means that many models do not need to specify the schema at all.
In the unit tests, I have updated the Galaxy class to use the default schema name instead of a custom one.

We may want to enhance the default implementation of AnyModel.schema() to be smarter about the rules for naming database tables. It might need to substitute underscores for other punctuation, or use all lowercase letters. But since the user can always override the default implementation, this is not critical.

I have also changed the examples in the unit tests to use `class func` rather than `static`, because `static` implies `final` and users may not want their model classes to be final. While static is fine in the unit tests, users who copy this into their own code may not realize that static implies final, and then get stuck when they try to subclass their own model classes.

### DIFF
--- a/Sources/FluentBenchmark/City.swift
+++ b/Sources/FluentBenchmark/City.swift
@@ -1,7 +1,7 @@
 import FluentKit
 
 public final class City: Model {
-    public static let schema = "cities"
+    public class func schema() -> String { "cities" }
 
     @ID(key: "id")
     public var id: Int?
--- a/Sources/FluentBenchmark/FluentBenchmarker+CustomID.swift
+++ b/Sources/FluentBenchmark/FluentBenchmarker+CustomID.swift
@@ -18,7 +18,7 @@ extension FluentBenchmarker {
 
 
 private final class Foo: Model {
-    static let schema = "foos"
+    class func schema() -> String { "foos" }
 
     @ID(key: "id")
     var id: UUID?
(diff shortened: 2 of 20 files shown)


samalone:
I have branches of the fluent-mysql-driver and fluent-postgres-driver repositories that fix the unit test failures above, but I don't know the best way to submit pull requests that need to be coordinated across multiple repositories.

calebkleveter:
@samalone The static property can't be stored on generic types, but it can be computed. This is perfectly valid:

```swift
final class ModelWrapper<M>: Model where M: Model {
    static var schema: String { return "\(M.schema)_wrapper" }

    // Everything else...
}
```

samalone:
You are correct. Somehow I missed this solution in my attempts to solve the problem.
I withdraw my pull request.

PR #214, opened 2020-03-17
Brute-force past a weird SPM issue in the Mongo test workflow

gwynne:
Running the tests in a Swift 5.2 container, even locally, fails out with the dreaded `Package.resolved` badly out of date error. Deleting it and letting package resolution re-run works around the problem without losing the revision pin on `fluent-kit`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,9 @@ jobs:
       working-directory: ./
     - run: swift package edit fluent-kit --revision ${{ github.sha }}
       working-directory: ./fluent-mongo-driver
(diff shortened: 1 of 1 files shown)


MrLotU:
Are we sure this only affects tests?

gwynne:
> Are we sure this only affects tests?

As far as I can tell, not only does it only affect tests, it only _specifically_ affects the Mongo tests, and them not 100% of the time. I mean, my guess is it's a corner-case bug in the new SPM resolver (you know, the one that actually gives useful errors half the time!) that'll get resolved in a few more nightlies anyhow.

PR #219, opened 2020-03-25
Support All Collection Types for Relation Property Wrappers

calebkleveter:
Right now we simply added support for `Set` along with `Array`. Hopefully, I can get a clean implementation that allows all `Collection` types, or if not that, all `ExpressibleByArrayLiteral` types.

### DIFF
--- a/Sources/FluentKit/Properties/Children.swift
+++ b/Sources/FluentKit/Properties/Children.swift
@@ -1,11 +1,11 @@
 extension Model {
-    public typealias Children<To> = ChildrenProperty<Self, To>
-        where To: FluentKit.Model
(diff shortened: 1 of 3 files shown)


calebkleveter:
Generally decided that this feature doesn't really make sense as you can't really enforce this on the database side, at least not easily.

https://discordapp.com/channels/431917998102675485/438857881970933780/692460562499502170

calebkleveter:
From @tanner0101:

> I think Set + children is worth exploring but I think it would require considering the feature as a whole. Including how it interacts with migrations and how the db could enforce this. In terms of impl, I'd worry about leaking impl details (and also generic bloat) and probably want @UniqueChildren or something similar.

PR #221, opened 2020-03-26
Custom Timestamp Formats

calebkleveter:
Adds a `format` parameter to the `Timestamp` property wrapper, allowing you to pass in a custom formatter that will be used to convert the stored `Date` to a codable type before it is store in the database. By default, the date will still be converted to the database-specific date type.

```swift
@Timestamp(key: "created_at", on: .create, format: .iso8601)
```


calebkleveter:
@tanner0101 This can come in very handy when someone wants to store a value differently then is implemented in the `<SQLDialect>DataConvertible` protocols. For example if they have a `Field` property that holds a date that isn't a timestamp, and they want that to be ISO 8601 also instead of a UNIX timestamp. The other type that I can imagine people messing with is `Decimal`, though they might come up with a plethora of other ones that they want to mess with.

In the end, it all comes down to #WhyNot?

tanner0101:
> In the end, it all comes down to #WhyNot?

More code = more problems. 

If you can show some strong examples of things that are difficult to do now that this feature would fix, that will help. Otherwise I'd prefer to start out with doing things as simply and minimally as possible. We can always add an abstraction in a minor release. We can't remove one though.

calebkleveter:
@tanner0101 I've removed everything related to value conversion from the `Field` and `OptionalField` properties.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -130,3 +171,31 @@ extension Schema {
         query.filters.append(.group([isNull, isFuture], .or))
     }
 }
+
+
+public protocol TimestampFormatter {
+    func string(from date: Date) -> String
```
tanner0101: Would it make sense to return `Encodable` here? For example `TimestampFormat.unix` would want to return an `Int`.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -130,3 +171,31 @@ extension Schema {
...
+    func date(from string: String) -> Date?
+}
+
+extension DateFormatter: TimestampFormatter { }
+extension ISO8601DateFormatter: TimestampFormatter { }
+
+public enum TimestampFormat {
```
tanner0101: Could we use an enum-like struct here instead to keep the cases internal? That would allow us to refactor this in the future if needed.

tanner0101: One potential refactor we may wanna do is caching thread-specific formatters for example.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -9,13 +20,14 @@ public enum TimestampTrigger {
 }
 
 @propertyWrapper
-public final class TimestampProperty<Model>
-    where Model: FluentKit.Fields
+public final class TimestampProperty<Model, Formatter>
```
MrLotU: Why is it generic over `Model` where model is of type `Fields`. Shouldn't that be renamed? Kind of confusing :)

calebkleveter: Pass that query to @tanner0101. That's what he did with all the other property wrappers, so I simply followed suit.

MrLotU: Yeah, figured as much. CC @tanner0101 might be good to check in a separate PR, non blocking for this one :)

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -1,5 +1,16 @@
+import class Foundation.ISO8601DateFormatter
+import class Foundation.DateFormatter
+import class NIO.ThreadSpecificVariable
+
 extension Fields {
-    public typealias Timestamp = TimestampProperty<Self>
+    public typealias Timestamp<Formatter> = TimestampProperty<Self, Formatter>
```
tanner0101: Adding a new generic parameter to a public type is a breaking change. For example:

```swift
import FluentKit

func foo(_ timestamp: User.Timestamp) {
    print(timestamp.key)
}
```

This would break and `foo` would need to become generic. 

```swift
func foo<Formatter>(_ timestamp: User.Timestamp<Formatter>) { ... }
```

calebkleveter: We haven't officially released FluentKit, so that's technically not an issue. 😁At the same time, I agree, we should avoid this. I'll see if I can get around that.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -74,33 +71,34 @@ extension TimestampProperty: AnyProperty {
...
+            self.field.outputValue = date
         } catch {
             throw FluentError.invalidField(name: self.field.key.description, valueType: Value.self, error: error)
         }
     }
 
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
-        try container.encode(self.wrappedValue.flatMap(self.formatter.timestamp(from:)))
+        try self.formatter.encode(date: self.value, to: &container)
```
MrLotU: Why not encode to the `Encoder`. No need for the inout that way.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -74,33 +71,34 @@ extension TimestampProperty: AnyProperty {
...
     }
 
     public func decode(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
         if container.decodeNil() {
             self.field.value = nil
         } else {
-            let string = try container.decode(Formatter.Timestamp.self)
-            self.field.inputValue = .bind(self.formatter.date(from: string))
+            let date = try self.formatter.decode(from: container)
```
MrLotU: Same here, but with `Decoder` :)

Sources/FluentKit/Properties/Timestamp.swift
calebkleveter: I wasn't sure if it would still work correctly. We'll wait on CI and see. 😄

calebkleveter:
@tanner0101 Are you happy now? 😉

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -130,3 +156,103 @@ extension Schema {
...
+        return new
+    }
+
+    private static var current: ThreadSpecificVariable<TimestampFormatterCache> = .init()
+
+    var cache: [String: AnyTimestampFormatter] = [:]
+}
+
+
+public struct TimestampFormat<Formatter> where Formatter: TimestampFormatter {
```
tanner0101: What if this held `AnyTimestampFormatter` instead of a generic one? That would let us make a lot of these types private.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -61,19 +71,35 @@ extension TimestampProperty: AnyProperty {
     }
     
     public func input(to input: inout DatabaseInput) {
-        self.field.input(to: &input)
+        let timestamp = self.value.map { date in self.formatter.anyTimestamp(from: date) } ?? Optional<Date>.none
```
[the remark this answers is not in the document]
MrLotU: I think `.bind(Date?.none)` aka `.bind(nil)` should become `DatabaseQuery.Value.null`, right?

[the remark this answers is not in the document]
calebkleveter: Apparently you can't. I just tried changing that and it made the CI fail:

```
2020-03-31T20:10:40.8647924Z 2020-03-31T20:10:40+0000 debug: database-id=sqlite INSERT INTO "users" ("id", "name", "deletedAt") VALUES (?, ?, NULL) ["19B01F27-6EFB-4CFA-876F-452ACC1FCC4C", "B"]
2020-03-31T20:10:40.8659561Z Fatal error: Invalid input type: nil: file /__w/fluent-kit/fluent-kit/fluent-sqlite-driver/Packages/fluent-kit/Sources/FluentKit/Model/Model+CRUD.swift, line 185
```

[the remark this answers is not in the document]
calebkleveter: This applies to all the drivers.

tanner0101:
These changes are now available in [1.0.0-rc.1.13](https://github.com/vapor/fluent-kit/releases/tag/1.0.0-rc.1.13)

PR #248, opened 2020-04-15
Query builder timestamps + soft delete

mcdappdev:
Adds a new `force` parameter to `QueryBuilder` `delete` calls:
```swift
MyModel.query(on: db).filter(\.$name == "Vapor").delete(force: true)
```
⚠️ The default behavior of `delete` on `QueryBuilder` has changed. If the model is soft deletable the query builder will now honor this instead of force deleting.

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder.swift
@@ -4,13 +4,13 @@ public final class QueryBuilder<Model>
     where Model: FluentKit.Model
 {
     public var query: DatabaseQuery
-    
+
     public let database: Database
     internal var includeDeleted: Bool
     internal var shouldForceDelete: Bool
     internal var models: [Schema.Type]
     public var eagerLoaders: [AnyEagerLoader]
-    
+
     public convenience init(database: Database) {
         self.init(
             query: .init(schema: Model.schema),
@@ -18,7 +18,7 @@ public final class QueryBuilder<Model>
             models: [Model.self]
         )
     }
-    
+
     private init(
         query: DatabaseQuery,
         database: Database,
@@ -41,7 +41,7 @@ public final class QueryBuilder<Model>
             self.query.customIDKey = idKey
         }
     }
-    
+
     public func copy() -> QueryBuilder<Model> {
         .init(
             query: self.query,
@@ -52,7 +52,7 @@ public final class QueryBuilder<Model>
             shouldForceDelete: self.shouldForceDelete
         )
     }
-    
+
     // MARK: Fields
 
     public func field<Field>(_ field: KeyPath<Model, Field>) -> Self
@@ -60,62 +60,62 @@ public final class QueryBuilder<Model>
     {
         self.field(Model.self, field)
     }
-    
+
     public func field<Joined, Field>(_ joined: Joined.Type, _ field: KeyPath<Joined, Field>) -> Self
         where Joined: Schema, Field: FieldProtocol, Field.Model == Joined
     {
         self.query.fields.append(.path(Joined.path(for: field), schema: Joined.schema))
         return self
     }
-    
+
     // MARK: Soft Delete
-    
+
     public func withDeleted() -> Self {
         self.includeDeleted = true
         return self
     }
-    
+
     // MARK: Actions
-    
+
     public func create() -> EventLoopFuture<Void> {
         self.query.action = .create
         return self.run()
     }
-    
+
     public func update() -> EventLoopFuture<Void> {
         self.query.action = .update
         return self.run()
     }
-    
+
     public func delete(force: Bool = false) -> EventLoopFuture<Void> {
         self.shouldForceDelete = force
         self.query.action = .delete
         return self.run()
     }
-    
+
     // MARK: Limit
-    
+
     public func limit(_ count: Int) -> Self {
         self.query.limits.append(.count(count))
         return self
     }
-    
+
     // MARK: Offset
-    
+
     public func offset(_ count: Int) -> Self {
         self.query.offsets.append(.count(count))
         return self
     }
-    
+
     // MARK: Unqiue
-    
+
     public func unique() -> Self {
         self.query.isUnique = true
         return self
     }
-    
+
     // MARK: Fetch
-    
+
     public func chunk(max: Int, closure: @escaping ([Result<Model, Error>]) -> ()) -> EventLoopFuture<Void> {
         var partial: [Result<Model, Error>] = []
         partial.reserveCapacity(max)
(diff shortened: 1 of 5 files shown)


Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -229,12 +234,29 @@ public final class QueryBuilder<Model>
...
+            }
+            
+            // If we don't want to force delete, just touch
+            // the timestamp model. We also change it to an
+            // update query. If it is a force delete, still
+            // touch the timestamps
+            if !self.forceDelete, let _ = initializedModel.deletedTimestamp {
+                initializedModel.touchTimestamps(.update, .delete)
+                query.action = .update
+                query.input = [.dictionary(initializedModel.input.values)]
```
tanner0101, line 254: I'm not sure this makes sense inside of `self.models`. I don't think Fluent supports deleting with joins yet or if the underlying DBs even support that. I think you are safe to assume here that the only model we care about during `delete` is the generic `Model`.

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -229,12 +234,29 @@ public final class QueryBuilder<Model>
...
+            } else {
+                switch query.action {
+                    case .create: initializedModel.touchTimestamps(.create, .update)
```
tanner0101, line 257: I don't think this will actually do anything. We should have some tests for it to be sure. You'll need to add the timestamp fields the existing values in `query.input`. This needs to work for both normal and batch saves. I'm also realizing now that once this is added here, we should be able to remove timestamp logic from the `Model+CRUD` file since it will be handled entirely here. That's a lot better 👍

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -38,34 +41,35 @@ public final class QueryBuilder<Model>
...
-            includeDeleted: self.includeDeleted
+            includeDeleted: self.includeDeleted,
+            forceDelete: self.forceDelete
         )
     }
-
+    
     // MARK: Fields
-
+
```
tanner0101, line 57: Can you trim whitespace lines? 
Screen Shot 2020-04-20 at 3 29 26 PM
(under "while editing")

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -134,11 +139,11 @@ public final class QueryBuilder<Model>
             .all()
             .map { $0.first }
     }
-
+    
     public func all<Field>(_ key: KeyPath<Model, Field>) -> EventLoopFuture<[Field.Value]>
         where
-            Field: FieldProtocol,
-            Field.Model == Model
+        Field: FieldProtocol,
```
tanner0101, line 145: Indentation

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -148,15 +153,15 @@ public final class QueryBuilder<Model>
...
+    
     public func all<Joined, Field>(
         _ joined: Joined.Type,
         _ field: KeyPath<Joined, Field>
     ) -> EventLoopFuture<[Field.Value]>
         where
-            Joined: Schema,
-            Field: FieldProtocol,
-            Field.Model == Joined
+        Joined: Schema,
```
tanner0101, line 162: Indentation

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -83,27 +87,28 @@ public final class QueryBuilder<Model>
         return self.run()
     }
     
-    public func delete() -> EventLoopFuture<Void> {
+    public func delete(force: Bool = false) -> EventLoopFuture<Void> {
+        self.forceDelete = Model.init().deletedTimestamp == nil ? true : force
```
tanner0101, line 91: I think this should be something like `shouldForceDelete`, then we do the actual check for `deletedTimestamp` in the run method. As is, having `forceDelete = false` on a model without timestamps is an invalid state that we could make impossible.

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -238,8 +243,46 @@ public final class QueryBuilder<Model>
...
+            if !self.forceDelete {
+                model.touchTimestamps(.delete, .update)
+                query.action = .update
+                query.input = [.dictionary(model.input.values)]
+            }
+        case .create:
+            var data: [DatabaseQuery.Value] = []
+            
+            for case .dictionary(var nested) in query.input {
+                if let createTimestamp = Model().timestamps.filter({ $0.trigger == .create }).first {
```
tanner0101, line 258: There's no limit to how many `on: .create` timestamps a model can have so doing `first` here is not technically correct.

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -238,8 +243,46 @@ public final class QueryBuilder<Model>
...
+                }
+                data.append(.dictionary(nested))
+            }
+            
+            query.input = data
+        case .update:
+            var data: [DatabaseQuery.Value] = []
+            
+            for case .dictionary(var nested) in query.input {
+                if let updateTimestamp = Model().timestamps.filter({ $0.trigger == .update }).first {
```
tanner0101, line 272: We could probably factor this out into a method since it's identical between both the create/update actions and on create/update timestamps.

Sources/FluentKit/Query/Builder/QueryBuilder.swift
mcdappdev: 🤯 I had no idea this setting was a thing

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -237,9 +242,41 @@ public final class QueryBuilder<Model>
...
+            }
+            
+            query.input = data
+        default:
+            break
+        }
         
         self.database.logger.info("\(self.query)")
-
+
```
tanner0101, line 279: Whitespace.

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -219,7 +224,7 @@ public final class QueryBuilder<Model>
         // make a copy of this query before mutating it
         // so that run can be called multiple times
         var query = self.query
-
+
```
tanner0101, line 227: Whitespace (above this line, too)

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -256,4 +293,15 @@ public final class QueryBuilder<Model>
         }
         return done
     }
+
+    private func addTimestamps(triggers: [TimestampTrigger], nested: inout [FieldKey: DatabaseQuery.Value]) {
+        let timestamps = Model().timestamps.filter { triggers.contains($0.trigger) }
+
+        for timestamp in timestamps {
+            let path = timestamp.path.first!
```
tanner0101, line 301: This will fail if the `@Timestamp` appears in an `@Group`. We should either give a clearer fatalError that that's not supported yet or properly implement this and add a test for it.

PR #255, opened 2020-04-22
Add @OptionalGroup support

123FLO321:
This adds support for @OptionalGroup properties.

I added a test cases for OptionalGroup to both fluent-kit and fluent-benchmark (however this test is not added to drivers yet).

It appears to be working correctly locally, but still needs further testing especially in combination with other property wrappers.
(@OptionalGroup and @Group inside a @OptionalGroup worked fine)

Problems:
 - This implementation brakes cases where `@Field` is used instead of `@OptionalField`


Sources/FluentKit/Properties/Field.swift
```diff
@@ -77,15 +77,23 @@ extension FieldProperty: AnyProperty {
     public func output(from output: DatabaseOutput) throws {
         if output.contains([self.key]) {
             self.inputValue = nil
+
+            // decode Value as Optional in case we are nested inside @OptionalGroup
+            // throw FluentError.unexceptedNil if the value is nil
+            let outputValue: Value?
             do {
-                self.outputValue = try output.decode(self.key, as: Value.self)
+                outputValue = try output.decode(self.key, as: Value?.self)
```
tanner0101, line 85: I'm not sure how decoding an `@OptionalGroup` will work. If the group can have `@OptionalField`s or nested `@OptionalGroup`s inside of it, then how would you tell whether or not the group is optional? 

For example, take this `Fields`:

```
struct Foo: Fields {
    @OptionalField var baz: String?
    @Field var bar: Int
    @OptionalGroup var qux: Qux
}

@OptionalGroup var foo: Foo?
```

When decoding this group, it seems like we should check whether `bar` exists or not to determine whether `foo` is nil. How do we express this in code though. Also, what happens if `Fields` does not contain any non-optional fields?

123FLO321: Decoding is done from deepest outwards. And since we don’t re-throw the error this only affects the next OptionalGroup when an nil @Field is found. 
Meaning only OptionalGroups that actually should be nil get set to nil. 

I tried nesting multiple OptionalGroups and seems to work just fine. 

If `Fields` contain no required fields we cannot determine if OptionalGroup should be nil or not resulting it in never being nil. (which is fine i guess. if you don’t have required fields why would one want to use @OptionalGroup regardless?)

123FLO321:
@tanner0101 `@Field` with Optional Values now work as expected again.

Sources/FluentKit/Properties/OptionalGroup.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    public func input(to input: inout DatabaseInput, chain: [AnyProperty]) {
+        if let values = self.value?.input(chain: chain+[self]).values, !values.isEmpty  {
+            input.values[self.key] = .dictionary(values)
+        }
+    }
+
+    public func output(from output: DatabaseOutput, chain: [AnyProperty]) throws {
+        do {
+            try self.value?.output(from: output.nested(self.key), chain: chain+[self])
```
tanner0101, line 64: > If Fields contain no required fields we cannot determine if OptionalGroup should be nil or not resulting it in never being nil. 

This feels like it has the potential to cause problems or confusion. For example, if you have a model with these fields:

```swift
struct Foo: Fields {
    @OptionalField var bar: Int?
    @OptionalField var baz: String?
}

@OptionalGroup var foo: Foo?
```

Say you save a new model with `foo = `nil`:

```swift
let a = Model()
a.foo = nil
try a.save(on: db).wait()
```

When you fetch this model, `foo` will no longer be `nil` because all of `Foo`'s fields are optional so no "unexpectedNil" error was thrown. 

```swift
let b = try Model.find(a.id!, on: db).wait()
print(b.foo) // Expected nil, but it is Foo
print(b.foo?.bar) // nil
print(b.foo?.baz) // nil
```

Sources/FluentKit/Properties/OptionalGroup.swift
```diff
@@ -0,0 +1,82 @@
...
+        } catch FluentError.unexceptedNil(_) {
```
tanner0101, line 65:
```suggestion
        } catch FluentError.unexpectedNil(_) {
```

Sources/FluentKit/Properties/NestedProperty.swift
```diff
@@ -43,19 +43,19 @@ extension NestedProperty: AnyProperty {
         fatalError()
     }
 
-    public func input(to input: inout DatabaseInput) {
+    public func input(to input: inout DatabaseInput, chain: [AnyProperty]) {
```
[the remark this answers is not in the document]
123FLO321: I can't really thing of another way. Any ideas?

Sources/FluentKit/Properties/OptionalGroup.swift
123FLO321: I don't think thats confusing.
We can just point that out in the wiki.
Having an @OptionalGroup with all fields beeing optional doesn't make any sense design wise.

123FLO321:
> Another idea could be to require that @OptionalGroup have at least one non-optional property. That property could be used in place of an explicit <key>_is_present field. I'm not really a fan of this either though since its difficult to make that behavior clear to the user. It feels too magical.

I think this makes a lot of sense though.
@OptionalGroup really only makes sense if you have a mandatory field in it.
Else normal @Group would work just fine

> `@Group var foo: Foo?`

That doesn't work since in `@Group`the Type can't be optional.  

>  Is there a use case I'm missing here that is not possible without @OptionalGroup?

The reason I coded it is because I really needed it to clean up my code.
I have Models that have groups that are either there or not (and all of them have mostly mandatory fields).
And those cases is where it's really useful.
Having to check just the group (foo) instead each individual field for nil (so bar, baz) is much cleaner.
Cases where `@OptionalGroup` is used and all Fields are optional doesn't make any sense because you're essentially just nesting Optionals then. 
There `@Group` should be used instead.

Sources/FluentKit/Properties/NestedProperty.swift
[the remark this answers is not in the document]
123FLO321: Also that shouldn't brake anything since the protocol still has the old function (without chain) that defaults the chain to empty.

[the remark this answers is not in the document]
tanner0101: The protocol requirement has changed though which means any conformers will break. I think the solutions I recommended in the main comment of this review would not require this. I'm not sure what `chain` is doing actually.

tanner0101:
> Cases where @OptionalGroup is used and all Fields are optional doesn't make any sense because you're essentially just nesting Optionals then.

I think this case does have value. Swift lets you have an optional struct with all optional values inside of it. Whether or not the struct is optional has meaning. For example:

```swift
struct User {
    var name: String?
}
var user: User?

if let user = user { 
    if let name = user.name {
        // The user has a name.
    } else {
        // The user does not have a name.
    }
} else {
    // There is no user.
}
```

This example is kind of silly, but I hope you see the point. A structure with all `nil` values has different meaning than a top-level `nil`. 

I think the best way forward here is a separate field like `<key>_is_present` unless there is a better idea I'm missing.

Jeehut:
@123FLO321 @tanner0101 That's the state on this, looks like it's not being worked on since nearly half a year. I just came across this with a structure like this:

```swift
final class Task: Model, Content {
    static let schema: String = "tasks"

    @ID(key: .id)
    var id: UUID?

    @Field(key: "name")
    var name: String

    @OptionalField(key: "details")
    var details: String?

    @Group(key: "recurrence_interval")
    var recurrenceInterval: SmartInterval?

    init() {}

    init(name: String) {
        self.name = name
    }
}
```

I'd love to make the `recurrenceInterval` an optional here as I want tasks to be creatable without a recurrence interval, and most tasks won't have one. The alternative to create a separate model for `SmartInterval` seems overkill to me, as the smart interval is going to be "owned" by the `Task` and there's no reason for reusability or an ID or similar things. The `@Group` looks like the perfect use case here. But it's missing the `@OptionalGroup` support as it seems ...

gwynne:
Closing due to age; please feel free to open a new PR if this is still something you want to see in Fluent!

Issue #218, opened 2020-03-25
Relation Property Assignment


calebkleveter:
Currently, model properties that use the `Sibling`, `Children`, or `Parent` properties wrappers (and others), do not support having their values set and then automagically handling the extra inserts/updates into the database. This is a feature we would like to support in the future.

- `@Parent` and `@OptionalParent`: When you set the value and then create or update that property's model in the database, we should also check to see if the parent model exists. If it doesn't, we should insert it. This can easily be done by calling an `INSERT ON CONFLICt IGNORE` query on the parent model (if you are using a SQL database).
- `@Siblings`: When creating or updating the model with a siblings property, we should insert new sibling models into its table if there are no matches where the model ID and sibling ID match. This can easily be checked because the `@Siblings` property wrapper has access to the pivot table and the key paths to the two properties that are the foreign IDs. If we want to be smart we will also delete any pivot rows that have the current model's ID but not any of the attached sibling IDs.
- `@Children`: The implementation for this would act differently based on the type of the parent property wrapper that it connects through. If the child uses an `@Parent` property wrapper, the child will have to be deleted from the database, since a value is required in the parent column. If both the parent and child models are soft-deletable, then the child model could be soft-deleted and the parent column would still have the ID of the parent that it is no longer connected to. If the child uses an `@OptionalParent` property wrapper, then the parent column would simply bet set to `nil`.

PR #276, opened 2020-05-14
Remove setters from read-only wrappers

ffried:
I wondered why those setters are provided. If they aren't, the compiler will already tell users that those properties aren't writable, removing the need to fail at runtime.

However, there might be reasons for these setters to exist that I haven't seen. If so, feel free to close this.

### DIFF
--- a/Sources/FluentKit/Properties/Children.swift
+++ b/Sources/FluentKit/Properties/Children.swift
@@ -30,15 +30,10 @@ public final class ChildrenProperty<From, To>
     }
 
     public var wrappedValue: [To] {
-        get {
-            guard let value = self.value else {
-                fatalError("Children relation not eager loaded, use $ prefix to access: \(name)")
-            }
-            return value
+        guard let value = self.value else {
+            fatalError("Children relation not eager loaded, use $ prefix to access: \(name)")
         }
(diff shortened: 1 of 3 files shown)


tanner0101:
Supporting for setting the properties should be added eventually, see https://github.com/vapor/fluent-kit/issues/218. In order to do this as a non-breaking change in the future the properties must be settable so that users are required to define the properties as `var` on their models.

ffried:
I see. Btw, you can't declare properties as `let` when they have a property wrapper (even if the property wrapper is readonly as well):
image

So it would be a non-breaking change either way. But if work on making those writable is already underway, I'm fine with that as well.

tanner0101:
Hmm weird. That must have changed before Swift 5.2 final. What happens when you try to set a property wrapper with no setter?

ffried:
@tanner0101 This results in an error (as if you'd try to assign a new value to a normal `let` constant): 
image

It also errors out when trying to modify the visibility of the setter:
image

Though that last error was added in Swift 5.2 and wasn't there in Swift 5.1. So in Swift 5.1 you could apply that `private(set)` even though it was useless since the underscore variable is `private` anyways.

PR #290, opened 2020-05-29
Additional fixes for #278 (problematic usage of XCTUnwrap)

### LINKED ISSUE #278: Use of unresolved identifier XCTUnwrap
When generating a project using `swift package generate-xcodeproj` and one of the targets has a dependency on `FluentBenchmark`. Such target doesn't compile. This supposedly works when using `open Package.swift`.

Note: in some cases using `open Package.swift` is not possible. In my specific case I am stuck in a limbo because of https://github.com/mongodb/mongo-swift-driver/issues/387 / https://bugs.swift.org/browse/SR-12009 that require the use of `swift package generate-xcodeproj` as a workaround.

See error below:
Screenshot_2020-05-15_at_16-1 45 37

The workaround described in this SO [answer](https://stackoverflow.com/a/59625008) works. I had to manually add those search paths into the FluentBenchmark target of my project.

It would be nice, if `FluentBenchmark` could avoid using `XCTUnwrap` until all these issues are fixed.

gwynne:
Previous fix for #278 was incomplete. This PR removes all remaining usages. Resolves #278 (again).

No semver label; fixes for tests don't need a release.

### DIFF
--- a/Sources/FluentBenchmark/Tests/MigratorTests.swift
+++ b/Sources/FluentBenchmark/Tests/MigratorTests.swift
@@ -62,20 +62,16 @@ extension FluentBenchmarker {
             let ids = Array(self.databases.ids())
             let databaseID = (ids[0], ids[1])
 
-            let database1 = try XCTUnwrap(
-                self.databases.database(
-                    databaseID.0,
-                    logger: Logger(label: "codes.vapor.tests"),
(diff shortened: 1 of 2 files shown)


valeriomazzeo:
Amazing thanks 👍

gwynne:
I see I need to update this repo's CI, but in the meantime...

PR #291, opened 2020-05-29
CI: Test all the drivers on all the Swift images at once

gwynne:
This one may just be impossible to get working...


gwynne:
Wow, that actually didn't take as much fiddling around as I expected. Still too much, though.

.github/workflows/test.yml
```diff
@@ -1,120 +1,190 @@
-name: test
+# TODO: When we get the ability to reference workflows from other repos, this one should pull in the ones from
+# each driver instead of this incomplete reimplementation of each one.
+# See https://github.community/t/is-it-possible-to-use-the-same-workflow-in-different-repositories/17719/6
+
+name: Test Matrix
+
```
MrLotU, line 6: Space begone

MrLotU: Actually. Nevermind 🤦

PR #394, opened 2020-10-15
Add @Child property for One-to-one relations (#74)

taikoo:
add Sources/FluentKit/Properties/Child.swift, edited from Sources/FluentKit/Properties/Children.swift:

see gist: https://gist.github.com/taikoo/5d92990f7baad57c9f2071aa908871df#file-child-patch

### DIFF
--- a/Sources/FluentKit/Properties/Child.swift
+++ b/Sources/FluentKit/Properties/Child.swift
@@ -1,0 +1,226 @@
+extension Model {
+    public typealias Child<To> = ChildProperty<Self, To>
+        where To: FluentKit.Model
+}
+
+// MARK: Type
+
+@propertyWrapper
+public final class ChildProperty<From, To>
+    where From: Model, To: Model
+{
+    public enum Key {
+        case required(KeyPath<To, To.Parent<From>>)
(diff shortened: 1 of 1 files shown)


Sources/FluentKit/Properties/Child.swift
```diff
@@ -0,0 +1,226 @@
...
+
+// MARK: Relation
+
+extension ChildProperty: Relation {
+    public var name: String {
+        "Child<\(From.self), \(To.self)>(for: \(self.parentKey))"
+    }
+
+    public func load(on database: Database) -> EventLoopFuture<Void> {
+        self.query(on: database).all().map {
```
0xTim, line 135: Why not just do `.first()` here instead?

0xTim (changes requested):
This will needs tests to test all the new functionality

Sources/FluentKit/Properties/Child.swift
taikoo: OK, I'll fix this.

kaishin:
Would be great to see this added. I'd have written tests but I am not familiar enough with fluent-kit to do so atm.

gwynne:
This functionality was implemented in #419

PR #405, opened 2020-11-11
More helpers for embedding raw SQL from the FluentKit layer

JaapWijnen:
This adds on to #400 and #402, allowing raw sql to be embedded in joins as well.
Example:
`query.join(.sql(embed: "LEFT JOIN (...) ON ... = ..."))

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Join.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Join.swift
@@ -178,6 +178,24 @@ extension QueryBuilder {
         ))
         return self
     }
+    
+    @discardableResult
+    public func join<Foreign>(
+        _ foreign: Foreign.Type,
+        _ join: DatabaseQuery.Join
+    ) -> Self
+        where Foreign: Schema
+    {
+        self.models.append(Foreign.self)
+        self.query.joins.append(join)
+        return self
+    }
+    
+    @discardableResult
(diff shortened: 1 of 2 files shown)


siemensikkema:
@JaapWijnen have you read Gwynne's comment here about why joins were not included? https://github.com/vapor/fluent-kit/pull/400#issuecomment-719318366

gwynne:
@JaapWijnen What @siemensikkema said.

JaapWijnen:
Ah I missed that, sorry. These proved very useful today. I'm not sure what doesn't work in this case though from @gwynne 's comment.

siemensikkema:
@gwynne I'd like to see if we can reconsider adding this (or at least a part of this) functionality after all. You mention the lack of variadics in Swift is not allowing us to specify the potentially multiple models being joined. I'd wager that in most cases you just need one extra model (this is covered by the first method in this PR) . So I'd like to take the pragmatic approach and making at least that case possible. We could even add overloads for 2 or 3 models if we feel generous. I think that would cover a lot of the use cases that people could need.
Furthermore, you might not even be interested in decoding the model you're joining at all. You might just want the "side effects" of how it influences the query. This is covered by the other methods in this PR.

JaapWijnen:
Yeah I mostly added it because I saw that it was done for other models, I mainly added it for the second method. I mean you're already doing raw SQL so you shouldn't expect too much to magically happen for you right?

PR #406, opened 2020-11-12
Pagination max per

kasperwelner:
**Objective:** 
Implement a maximum `per` value for pagination in order to prevent the possibility of overloading the server by requesting very large datasets.

**Discussion:**
The simplest solution I have been able to come up with is overloading the `paginate` function on QueryBuilder:

```
extension QueryBuilder {
    /// Returns a single `Page` out of the complete result set according to the supplied `PageRequest`.
    ///
    /// This method will first `count()` the result set, then request a subset of the results using `range()` and `all()`.
    ///
    /// - Parameters:
    ///     - request: Describes which page should be fetched.
    ///     - maxPer: If `per` value of supplied `PageRequest` exceeds this value an error will be thrown. Default is 100.
    ///
    /// - Returns: A single `Page` of the result set containing the requested items and page metadata.
    public func paginate(
        _ request: PageRequest,
        maxPer: Int = 100
        guard request.per <= maxPer else {
            return database.eventLoop.makeFailedFuture(FluentError.maxPerValueExceeded)
        }
        let count = self.count()
        let items = self.copy().range(request.start..<request.end).all()
        return items.and(count).map { (models, total) in
            Page(
                items: models,
                metadata: .init(
                    page: request.page,
                    per: request.per,
                    total: total
                )
            )
        }
    }
}
```

However, right now the limit is hardcoded and I am yet to find a good solution for making it globally configurable. Considerations:
- I would like to keep it in fluent-kit and not introduce any new dependencies
- fluent-kit is not dependent on Vapor and thus I can’t use any of the features offered by Vapor for setting global state
- Database has Context.Configuration that potentially could hold this value but it doesn’t really sit well with me to put it there

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -2,12 +2,18 @@ extension QueryBuilder {
     /// Returns a single `Page` out of the complete result set according to the supplied `PageRequest`.
     ///
     /// This method will first `count()` the result set, then request a subset of the results using `range()` and `all()`.
+    ///
     /// - Parameters:
     ///     - request: Describes which page should be fetched.
+    ///     - maxPer: If `per` value of supplied `PageRequest` exceeds this value an error will be thrown. Default is 100.
     /// - Returns: A single `Page` of the result set containing the requested items and page metadata.
     public func paginate(
-        _ request: PageRequest
+        _ request: PageRequest,
+        maxPer: Int = 100
     ) -> EventLoopFuture<Page<Model>> {
+        guard request.per <= maxPer else {
+            return database.eventLoop.makeFailedFuture(FluentError.maxPerValueExceeded)
+        }
         let count = self.count()
         let items = self.copy().range(request.start..<request.end).all()
         return items.and(count).map { (models, total) in
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -9,11 +9,12 @@ extension QueryBuilder {
     ///     
     /// - Returns: A single `Page` of the result set containing the requested items and page metadata.
     public func paginate(
-        _ request: PageRequest,
-        maxPer: Int = 100
+        _ request: PageRequest
     ) -> EventLoopFuture<Page<Model>> {
-        guard request.per <= maxPer else {
-            return database.eventLoop.makeFailedFuture(FluentError.maxPerValueExceeded)
+        if let maxPerPage = database.context.maxPerPage {
+            guard request.per <= maxPerPage else {
+                return database.eventLoop.makeFailedFuture(FluentError.maxPerPageValueExceeded)
+            }
         }
         let count = self.count()
         let items = self.copy().range(request.start..<request.end).all()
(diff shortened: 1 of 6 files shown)


cweinberger:
My 2 cents:
- I think having this limit is absolutely useful and should be in! Great suggestion 👍 
- It feels wrong to have it in fluent-kit though, but that's because it already feels wrong that we have PageRequest defined in Fluent-Kit; I would suggest to move the PageRequest to Vapor Core and then use the usual options to make things configurable

kasperwelner:
I definitely agree with this. 

> My 2 cents:
> 
> * I think having this limit is absolutely useful and should be in! Great suggestion 👍
> * It feels wrong to have it in fluent-kit though, but that's because it already feels wrong that we have PageRequest defined in Fluent-Kit; I would suggest to move the PageRequest to Vapor Core and then use the usual options to make things configurable
> 
> Why have it in Vapor? Because PageRequest is not only useful for paginating database results but also for anything other than that (e.g. collections retrieved from local storage or other apis)

I agree. It does have the implication that `QueryBuilder.paginate()`will have to be moved to the Fluent package in order to keep fluent-kit free of a dependency on Vapor. 

I'll draft up something.

cweinberger:
FluentKit could still have a representation of Page/Pagination which can be initialized from the Vapor counterpart.

kasperwelner:
> FluentKit could still have a representation of Page/Pagination which can be initialized from the Vapor counterpart.

The paginate function could also simply take the page and per arguments directly

siemensikkema:
@kasperwelner I think we can do something similar to how QueryHistory works. We should be able to do something like:
```swift
// configure application default:
application.fluent.pagination.maxPerPage = 100 // optional value, default = nil means no limit

// optionally override for a request:
request.fluent.pagination.maxPerPage = 99
```
This value can be communicated from `Application.Fluent`/``Request.Fluent` → `Application.Databases` (the `database` function) → `DatabaseContext` (just like `QueryHistory`) so it can be accessible to `QueryBuilder` through `database.context.maxPerPage`. It's a bit of a detour but I think it makes for a good API and allows you to set a max limit to an existing project with 1 line of code.

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
```diff
@@ -2,12 +2,20 @@ extension QueryBuilder {
     /// Returns a single `Page` out of the complete result set according to the supplied `PageRequest`.
     ///
     /// This method will first `count()` the result set, then request a subset of the results using `range()` and `all()`.
+    ///
     /// - Parameters:
     ///     - request: Describes which page should be fetched.
+    ///     - maxPer: If `per` value of supplied `PageRequest` exceeds this value an error will be thrown. Default is 100.
+    ///
```
siemensikkema, line 9: This is outdated.

siemensikkema (commented):
@pulpn0ir Looks good! 🤩 Only thing missing is a convenient API to set the limit. I think that should go through the `.fluent` namespace on both `Application` and `Request`. See also my previous comment on this PR.

siemensikkema:
@pulpn0ir By the way, you should be able to fix the failing mac test by changing line 107 from the test workflow file to:
```
        uses: maxim-lobanov/setup-xcode@v1
```

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
pulpn0ir: Removed

siemensikkema:
Closing in favor of a new PR to be opened by @pulpn0ir who is taking this one over. (I couldn't find out how to transfer a PR)

PR #412, opened 2021-01-12
Add page size limit

pulpn0ir:
Allows setting a page size limit when creating a `DatabaseContext` to prevent server overloads.

If the requested page size exceeds the limit, the response page is capped to the defined limit. The default value is `nil`, allowing for an unlimited page size.

This PR relates to https://github.com/vapor/fluent/pull/715

### DIFF
--- a/Sources/FluentKit/Database/Database.swift
+++ b/Sources/FluentKit/Database/Database.swift
@@ -46,8 +46,8 @@ extension Database {
         self.context.history
     }
 
-    public var maxPerPage: Int? {
-        self.context.maxPerPage
+    public var pageSizeLimit: Int? {
+        self.context.pageSizeLimit
     }
 }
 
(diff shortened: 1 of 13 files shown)


siemensikkema:
This PR is the continuation of #406

Sources/FluentKit/Database/Database.swift
```diff
@@ -45,6 +45,10 @@ extension Database {
     public var history: QueryHistory? {
         self.context.history
     }
+
+    public var maxPerPage: Int? {
```
calebkleveter, line 49: The name is a little ambiguous. Something like `pageSizeLimit` would be more self-explanatory.

siemensikkema: @calebkleveter I can get on board with that. @pulpn0ir what about you?

pulpn0ir: Fully agree. I updated both PRs.

0xTim (approved):

Joannis:
@0xTim any reason this wasn't merged (yet)?

0xTim:
@siemensikkema I was waiting for you to give the final OK for this, this good to merge? Don't merge just yet if so, just approve. I'll be sorting out the release bot in a couple of hours so we can wait for that

Joannis:
After a small discussion with @0xTim it seems he wants at least @gwynne or @siemensikkema to look into this as well.

gwynne (changes requested):
Looks mostly fine to me, but I'd like the unit tests to include a test for a page other than the first (in particular, a test that pages after the first are at the correct offset relative to the given limits).

siemensikkema:
@gwynne good idea. I changed the test slightly to ensure this is covered. I chose to reduce the page limit so I could request page 2 while still getting a full page. The other choice I could have made would have been to add one more row to the test output.

Issue #414, opened 2021-01-28
`pg_enum` table is not adding all the cases added through single enum update migration


susheel-clearcals:
### Describe the bug

On trying to update an enum as shown below only the last case is getting added to the `pg_enum` table. In this case, the `pg_enum` only adds `teaCup` case. As a result of this the tables that refer to this enum as a datatype are not picking up all the update cases.

```swift
database.enum(SchemaConstants.recipeUnit)
                .case("piece")
                .case("glass")
                .case("mug")
                .case("teaCup")
                .update()
                .transform(to: ())
```

### To Reproduce

1. Add package with following configuration 
```swift
.package(url: "https://github.com/vapor/vapor.git", from: "4.36.0"),
.package(url: "https://github.com/vapor/fluent.git", from: "4.0.0"),
```
2. Create an enum model and migration through Fluent.migration.
3. Create another migration with a sequence of new cases in the enum and update the same.

### Expected behavior

The `pg_enum` table should have all the new cases that are update through the migration.

### Environment
* Vapor Framework version: 4.0
* Vapor Toolbox version: 4.0
* OS version: Mac Big Sur
* Database: Postgres

siemensikkema:
@susheel-clearcals I wonder if this has anything to do with it: https://github.com/vapor/fluent-kit/pull/411. Could you try using FluentKit 1.10.2 explicitly to see if this still occurs there?

madsodgaard:
Pretty sure this issue is more related to a driver limitation. In Postgres it seems like you can only have one `ADD VALUE` inside `ALTER TYPE`. Therefore, only the last case is added in this example.

A workaround is just iterating over the cases you want to create, and creating each in its own query.

```swift
let newValues = ["val1", "val2"]

return newValues.map { value in
  database.enum("schema").case(value).update()
}
.flatten(on: db.eventLoop)
```

susheel-clearcals:
Agree with @madsodgaard. It is indeed related to `ALTER TYPE` supporting only one `ADD VALUE`. Satisfied with the work around. Closing the ticket.

fananek:
The reason is clear. I'm just wondering whether vapor should handle this under the hood or at least mention this behavior in documentation.

siemensikkema:
@fananek I agree that Vapor ideally should handle this but it seems that it's not only postgres that is affected but any database that uses `SQLAlterEnumBuilder` since that can only ever contain one case value. SQLDatabases need to run this builder once for each added case in order to fix the issue.

PR #416, opened 2021-02-04
Add test for adding multiple enum cases at once

siemensikkema:
This draft adds a test for adding multiple enum cases in a single update. This should show that the behavior reported in #414 is in fact not supported (yet). Fixing this will require updates to the individual database drivers.


siemensikkema:
It appears that only Postgres is affected which makes sense looking at the implementations since the other databases don't need an explicit migration for enums.

0xTim (approved):

PR #419, opened 2021-02-22
Add @OptionalChild property wrapper (one-to-one relation)

fananek:
Add new `@OptionalChild` property for one-to-one relations.

Example:
```swift
private final class Planet: Model {
    static let schema = "planets"

    @ID(key: .id)
    var id: UUID?

    @Field(key: "name")
    var name: String

    @OptionalChild(for: \.$planet)
    var governor: Governor

    init() { }

    init(id: IDValue? = nil, name: String) {
        self.id = id
        self.name = name
    }
}

private final class Governor: Model {
    static let schema = "governors"

    @ID(key: .id)
    var id: UUID?

    @Field(key: "name")
    var name: String

    @Parent(key: "planet_id")
    var planet: Planet

    init() { }

    init(id: IDValue? = nil, name: String, planetID: UUID) {
        self.id = id
        self.name = name
        self.$planet.id = planetID
    }
}
```

See test files `ChildTests.swift` and `Governor.swift` for more details.

This PR:
- Closes issue #74 
- Supersedes #394 (which seems to be inactive)


Sources/FluentKit/Properties/Child.swift
```diff
@@ -0,0 +1,241 @@
...
+        to.forEach {
+            switch self.parentKey {
+            case .required(let keyPath):
+                $0[keyPath: keyPath].id = id
+            case .optional(let keyPath):
+                $0[keyPath: keyPath].id = id
+            }
+        }
+        return to.create(on: database)
+    }
```
siemensikkema, line 77: This method kind of goes against the idea of there being a single child.

siemensikkema (commented):
Thanks for your PR! Am I correct in understanding that this in essence offers the same functionality as doing `.first` on `Children` property? This is not to downplay your work but just for my understanding. As in, it does not seem to offer any guarantees that are not more `Child`ren for a given parent model. That would still need to be enforced through a database constraint for instance.

Sources/FluentKit/Properties/Child.swift
fananek: Thanks for reviewing the code. You are absolutely right in both points.
- The method is a leftover I missed. I already pushed new commit to PR.
- Current implementation requires `.unique` database constraint to ensure `Child` uniqueness. I'm open to any ideas if there is a feasible way to check whether `Child` exists during creation.

0xTim:
This should allow you to define a single child property on the parent `var child: MyChild` rather than an array. This definitely simplifies several code paths instead of having to deal with an array

fananek:
@siemensikkema. @0xTim - Hi guys, is there anything else I can do in order to successfully merge and close this PR?

Sources/FluentKit/Properties/Child.swift
```diff
@@ -0,0 +1,226 @@
...
+            self.idValue = try output.decode(key, as: From.IDValue.self)
+        }
+    }
+}
+
+// MARK: Codable
+
+extension ChildProperty: AnyCodableProperty {
+    public func encode(to encoder: Encoder) throws {
+        if let rows = self.value {
```
0xTim, line 116: This should probably be a single row or renamed to `child` since it's no longer an array

0xTim:
This is great! Thanks for this. My only query is whether `@Child` should only be applied to 
optional properties. When you first create a `@Parent` there's no child so getting the model and eager loading may cause issues. Additionally, I'd like to see a test for `query(on:)` for Child just to show how it works - I'm assuming this requires a `.first()` to get the actual value?

0xTim:
I'm happy to defer to a lot of the community for this one though, but I'd like @gwynne to chime in as well

Sources/FluentKit/Properties/Child.swift
fananek: Good catch, thanks. New commit already in place.

0xTim:
Just to follow up from Discord - the consensus is that `@Child` should be restricted to optional values only

fananek:
@0xTim Hi Tim, the latest version is available for review. Hope I got it right.
- Adjusted Child implementation.
- `@Child` has been renamed to `@OptionalChild` so we keep naming consistent.
- There are few more `query(on: )` tests for different relations

Sources/FluentBenchmark/SolarSystem/Governor.swift
```diff
@@ -0,0 +1,70 @@
...
+        self.id = id
+        self.name = name
+        self.$planet.id = planetId
+    }
+}
+
+public struct GovernorMigration: Migration {
+    public func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("governors")
+            .field("id", .uuid, .identifier(auto: false))
```
gwynne, line 32: Add `.unique` and `.required` constraints here for completeness' sake.
```suggestion
            .field(.id, .uuid, .identifier(auto: false), .required, .unique)
```

Sources/FluentBenchmark/SolarSystem/Governor.swift
```diff
@@ -0,0 +1,70 @@
...
+                    governor = .init(name: "John Doe")
+                case "Earth":
+                    governor = .init(name: "Jane Doe")
+                default:
+                    governor = nil
+                }
+                guard let gov = governor else {
+                    return database.eventLoop.makeSucceededFuture(())
+                }
+                return planet.$governor.create(gov, on: database)
```
gwynne, line 62:
```suggestion
                default:
                    return database.eventLoop.future(())
                }
                return planet.$governor.create(governor!, on: database)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+    init(id: IDValue? = nil, name: String) {
+        self.id = id
+        self.name = name
+    }
+}
+
+private struct FooMigration: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("foos")
+            .field("id", .uuid, .identifier(auto: false))
```
gwynne, line 96:
```suggestion
            .field(.id, .uuid, .identifier(auto: false), .required, .unique)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+        self.id = id
+        self.bar = bar
+        self.$foo.id = fooID
+    }
+}
+
+private struct BarMigration: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("bars")
+            .field("id", .uuid, .identifier(auto: false))
```
gwynne, line 130:
```suggestion
            .field(.id, .uuid, .identifier(auto: false), .required, .unique)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+    init(id: IDValue? = nil, bar: Int, fooID: Foo.IDValue) {
+        self.id = id
+        self.bar = bar
+        self.$foo.id = fooID
+    }
+}
+
+private struct BarMigration: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("bars")
```
gwynne, line 129:
```suggestion
        database.schema(Bar.schema)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    init(id: IDValue? = nil, name: String) {
+        self.id = id
+        self.name = name
+    }
+}
+
+private struct FooMigration: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("foos")
```
gwynne, line 95:
```suggestion
        database.schema(Foo.schema)
```

Sources/FluentBenchmark/SolarSystem/Governor.swift
```diff
@@ -0,0 +1,70 @@
...
+        database.schema("governors")
+            .field("id", .uuid, .identifier(auto: false))
+            .field("name", .string, .required)
+            .field("planet_id", .uuid, .required, .references("planets", "id"))
+            .unique(on: "planet_id")
+            .create()
+    }
+
+    public func revert(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("governors").delete()
```
gwynne, line 40:
```suggestion
        database.schema(Governor.schema).delete()
```

Sources/FluentBenchmark/SolarSystem/Governor.swift
```diff
@@ -0,0 +1,70 @@
...
+    public init(id: IDValue? = nil, name: String, planetId: UUID) {
+        self.id = id
+        self.name = name
+        self.$planet.id = planetId
+    }
+}
+
+public struct GovernorMigration: Migration {
+    public func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("governors")
```
gwynne, line 31:
```suggestion
        database.schema(Governer.schema)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+            .field("id", .uuid, .identifier(auto: false))
+            .field("name", .string, .required)
+            .create()
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("foos").delete()
```
gwynne, line 102:
```suggestion
        database.schema(Foo.schema).delete()
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+        database.schema("bars")
+            .field("id", .uuid, .identifier(auto: false))
+            .field("bar", .int, .required)
+            .field("foo_id", .uuid, .required)
+            .unique(on: "foo_id")
+            .create()
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("bars").delete()
```
gwynne, line 138:
```suggestion
        database.schema(Bar.schema).delete()
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+        self.id = id
+        self.baz = baz
+        self.$foo.id = fooID
+    }
+}
+
+private struct BazMigration: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("bazs")
+            .field("id", .uuid, .identifier(auto: false))
```
gwynne, line 166:
```suggestion
        database.schema(Baz,schema)
            .field(.id, .uuid, .identifier(auto: false), .required, .unique)
```

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,175 @@
...
+            .field("baz", .double, .required)
+            .field("foo_id", .uuid)
+            .create()
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("bazs").delete()
```
gwynne, line 173:
```suggestion
        database.schema(Baz.schema).delete()
```

gwynne (changes requested):
Looks great, most of my comments are minor nits. One request: Can we add a SQL-mode-only test for ensuring `@OptionalChild` also works with `@ID` columns that are not `UUID`s? (SQL mode can be detected either by the mechanism in `SQLTests.swift` of checking `database as? SQLDatabase`, or by one of my and Tanner's several inconsistent attempts to provide it as an input to the benchmarks, as in `SortTests.swift` 🤦‍♀️)

I'll also see if I can't get another at least incremental update to the CI done.

gwynne:
@fananek This is excellent work, thank you for doing this!

Sources/FluentBenchmark/Tests/ChildTests.swift
```diff
@@ -0,0 +1,285 @@
...
+private final class Game: Model {
+    static let schema = "games"
+
+    @ID(custom: .id, generatedBy: .database)
+    var id: Int?
+
+    @Field(key: "title")
+    var title: String
+
+    // It's a solitare game :P
```
gwynne, line 221: I lol'd at this - well played! 🤣

gwynne (approved):
I love it! Cut, print it, ship it!! 💯💯

gwynne:
@fananek One last item - the pull request's title and description are used by the bot to create the new release once the PR is merged; can you update the text accordingly? If you aren't sure what to change, I can take care of it, but I always try to give the original author first crack 🙂.

fananek:
@gwynne, @0xTim, @siemensikkema  Big thanks to all participants for reviews, suggestions and overall support with this PR!

Sources/FluentBenchmark/Tests/ChildTests.swift
fananek: Just a small `easter egg` comment. :)

tanner0101:
These changes are now available in [1.11.0](https://github.com/vapor/fluent-kit/releases/tag/1.11.0)

PR #429, opened 2021-04-28
Add async/await APIs to FluentKit

0xTim:
Adds support for async/await to FluentKit. This initially adds async APIS most public APIs that return `EventLoopFuture`. This PR also introduces a new `AsyncMigration` to provide an async API for migrations

This PR will be merged when async/await functionality lands in Swift and NIO and will be kept in the branch until then.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,6 +94,9 @@ jobs:
           - swiftlang/swift:nightly-5.3-bionic
           - swiftlang/swift:nightly-5.3-focal
           - swiftlang/swift:nightly-5.3-amazonlinux2
+          - swiftlang/swift:nightly-5.4-bionic
+          - swiftlang/swift:nightly-5.4-focal
+          - swiftlang/swift:nightly-5.4-amazonlinux2
     container: ${{ matrix.image }}
     steps:
       - name: Checkout code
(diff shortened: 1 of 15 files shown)


Sources/FluentKit/Concurrency/Database+Concurrency.swift
```diff
@@ -0,0 +1,15 @@
+#if compiler(>=5.5) && $AsyncAwait
+import _NIOConcurrency
+
+@available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+public extension Database {
+    func transaction<T>(_ closure: @escaping (Database) -> EventLoopFuture<T>) async throws -> T {
```
madsodgaard, line 6: Shouldn't the closure be `_ closure: @escaping (Database) async throws -> T` as well?

0xTim: Oops! Yep it should be. See #439

mxcl:
NIO now has sensible @availables for the async functions. I would immediately PR but there seems to be issues with Xcode 13 beta 2 that I’m trying to figure out.

0xTim:
@mxcl everything has now been updated

PR #432, opened 2021-05-18
Add conditional filtering to QueryBuilder for ModelAlias

### LINKED ISSUE #431: Filter joined ModelAlias(es) by shared properties
**Existing Problem** 
Fluent queries do not permit conditional filtering by properties shared among joined ModelAlias(es). Currently, you can only create a filter such as this on the originally queried model. For example, this is currently possible:

```swift
    let model = Team()
}
    let model = Team()
}
 
/// Find games that were popular enough for the crowd to be larger
/// than the number of players playing it.        
Match.query(on: self.database)
    /// You can currently conditionally filter the originally 
    /// queried Model type if types match (in this case,  
    /// crowd size and total players would be of type Int)
    .filter(\Match.$crowd, .greaterThan, \Match.$totalPlayers)
    .all()
```

**Proposed Solution**
Query builders would allow conditional filtering of joined ModelAliases. The call site would look something like:

```swift
Match.query(on: self.database)
    /// Here you can filter a ModelAlias according to properties shared with 
    /// another ModelAlias of the same original Model type.
    .filter(\HomeTeam.$score, .greaterThan, \AwayTeam.$score)
    .all()
```

To achieve this, I imagine a modification of `QueryBuilder+Filter.swift` whereby existing filter methods  are combined something similar to below to allow conditional filtering on Joined types:

```swift

    /// Please note this code is for demonstration only 
    /// and is not functional.
    @discardableResult
    public func filter<Left, Right>(
        _ lhsField: KeyPath<FirstModelAlias, Left>,
        _ method: DatabaseQuery.Filter.Method,
        _ rhsField: KeyPath<SecondModelAlias, Right>
    ) -> Self
        where Left: QueryableProperty,
            Left.Model == FirstModelAlias,
            Right: QueryableProperty,
            Right.Model == SecondModelAlias
    {
        self.filter(FirstModelAlias.path(for: lhsField), method, SecondModelAlias.path(for: rhsField))
    }
   ```

### PR DESCRIPTION
**Pull request by jerrycanco**
Adds support for QueryBuilder to filter joined ModelAlias(es) by similar properties to close #431  

Apologies in advance if I've mucked something up, I've never done a pull request before.

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Filter.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Filter.swift
@@ -43,6 +43,28 @@ extension QueryBuilder {
     {
         self.filter(Model.path(for: lhsField), method, Model.path(for: rhsField))
     }
+    
+    @discardableResult
+    public func filter<LeftJoined, LeftField, RightJoined, RightField>(
+        _ leftJoined: LeftJoined.Type,
+        _ leftField: KeyPath<LeftJoined, LeftField>,
+        _ method: DatabaseQuery.Filter.Method,
+        _ rightJoined: RightJoined.Type,
+        _ rightField: KeyPath<RightJoined, RightField>
+    ) -> Self
+    where LeftJoined: Schema,
+          RightJoined: Schema,
+          LeftField: QueryableProperty,
+          RightField: QueryableProperty,
+          LeftField.Model == LeftJoined,
+          RightField.Model == RightJoined
+    {
+        self.filter(
+            .path(LeftJoined.path(for: leftField), schema: LeftJoined.schemaOrAlias),
+            method,
+            .path(RightJoined.path(for: rightField), schema: RightJoined.schemaOrAlias)
+        )
+    }
 
     @discardableResult
     public func filter<Value>(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by 0xTim — APPROVED** [review]
LGTM - @gwynne any comments?

File: Sources/FluentKit/Query/Builder/QueryBuilder+Filter.swift [hunk 1]
```diff
@@ -43,6 +43,28 @@ extension QueryBuilder {
     {
         self.filter(Model.path(for: lhsField), method, Model.path(for: rhsField))
     }
+    
+    @discardableResult
+    public func filter<LeftJoined, LeftField, RightJoined, RightField>(
+        _ leftJoined: LeftJoined.Type,
+        _ leftField: KeyPath<LeftJoined, LeftField>,
+        _ method: DatabaseQuery.Filter.Method,
+        _ rightJoined: RightJoined.Type,
```
calebkleveter, line 52: Could we add a default value to the `leftJoined` and `rightJoined` parameters so we can use this syntax to call the method?

```swift
.filter(\FooAlias.$number, .equal, \BarAlias.$number)
```

gwynne:
Closing as resolved by #593

PR #433, opened 2021-05-24
Fix crash with negative `page` or `per` values

### LINKED ISSUE #434: paginate() with negative "per" argument causes application crash
### Describe the bug

Using the `paginate` function and combining it with a negative `per` query argument crashes the application. 

The paginate query arguments are decoded to `Int`'s and thus accepting negative values. The `per` value is then used in a range expression (in `QueryBuilder.paginate`), which causes the error (if `per` is negative).

### To Reproduce

In `routes.swift`:
```swift
app.get("planets") { req in
    Planet.query(on: req.db).paginate(for: req)
}
```

Send the a request with a negative `per` value as seen below:
```
curl -X GET -i -s 'http://localhost:8080/planets?page=1&per=-1'
```

The result is a crashed server with the following error message:
```
Fatal error: Range requires lowerBound <= upperBound: file Swift/x86_64-apple-macos.swiftinterface, line 14990
2021-05-22 19:23:25.638617+0200 Run[97014:17087570] Fatal error: Range requires lowerBound <= upperBound: file Swift/x86_64-apple-macos.swiftinterface, line 14990
```

### Expected behavior

The code should be able to handle negative `per` values. Expected behavior is probably to treat negative values as zero. Another solution would be to not accept faulty arguments (which then also applies to `page`, which treats values <= 0 as 1.

### Environment

framework: 4.45.4
toolbox: 18.3.3
macOS: 11.2.3

0xTim:
This fixes an issue where specifying a negative page number of number of items in a pagination request caused FluentKit to crash. Resolves #434

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -10,10 +10,13 @@ extension QueryBuilder {
         _ request: PageRequest
     ) -> EventLoopFuture<Page<Model>> {
         let trimmedRequest: PageRequest = {
-            guard let pageSizeLimit = database.context.pageSizeLimit else { return request }
+            guard let pageSizeLimit = database.context.pageSizeLimit else {
+                return .init(page: Swift.max(request.page, 1), per: Swift.max(request.per, 1))
+
+            }
             return .init(
-                page: request.page,
-                per: Swift.min(request.per, pageSizeLimit)
+                page: Swift.max(request.page, 1),
+                per: Swift.max(Swift.min(request.per, pageSizeLimit), 1)
             )
         }()
         let count = self.count()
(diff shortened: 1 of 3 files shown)


Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
```diff
@@ -10,10 +10,13 @@ extension QueryBuilder {
         _ request: PageRequest
     ) -> EventLoopFuture<Page<Model>> {
         let trimmedRequest: PageRequest = {
-            guard let pageSizeLimit = database.context.pageSizeLimit else { return request }
+            guard let pageSizeLimit = database.context.pageSizeLimit else {
+                return .init(page: Swift.max(request.page, 1), per: Swift.max(request.per, 1))
+
```
siemensikkema, line 15: nit: newline

siemensikkema (approved):
That should do it! (Just found a rogue newline)

PR #439, opened 2021-06-11
Add async for closure in `Database.transaction async`

mxcl:
Probably this is what we want. If so, I can update the other function in this file too.

To be clear this allows:

```swift
let model = try await db.transaction { db -> Model in
     let foo = try await Model.query(on: db).first()
     foo.bar = 1
    try await foo.update(on: db)
}
```

Rather than:

```swift
let model = try await db.transaction { db in
     Model.query(on: db).first().flatMap { … }
}
```


0xTim (approved):
Oops my bad. Yep definitely want it to be like this. If you could update the other function that would be awesome

PR #440, opened 2021-07-06
Remove `Codable` restriction on `Page.map`

siemensikkema:
Allow `map`ing items in a `Page` to non-Codable types (#440).

Since https://github.com/vapor/fluent-kit/pull/435 this restriction is no longer necessary.


0xTim (approved):
LGTM

Issue #446, opened 2021-08-01
Async/await ModelMiddleware doesn't work


kevinrenskers:
I am working with the async-await branches of both Vapor and FluentKit, and overall it's all working perfectly fine and really improves the DX of working with Vapor. Well done!

However, the ModelMiddelware isn't working properly.

Old-fashioned is working fine:

``` swift
struct CampaignModelMiddleware: ModelMiddleware {
  func update(model: Campaign, on db: Database, next: AnyModelResponder) -> EventLoopFuture<Void> {
    return next.update(model, on: db).map {
      print ("Campaign \(model.name) was updated")
    }
  }
}
```

But the async/await version is not:

``` swift
struct CampaignModelMiddleware: ModelMiddleware {
  func update(model: Campaign, on db: Database, next: AnyModelResponder) async throws {
    try await next.update(model, on: db).get()
    print("Campaign \(model.name) was updated")
  }
}
```

It's never executed. I think the problem is that while the protocol methods and their default implementation are given, those methods are never actually called from within the [`ModelMiddleware.handle` method](https://github.com/vapor/fluent-kit/blob/async-await/Sources/FluentKit/Middleware/ModelMiddleware.swift#L38).

elidydev:
I would like to take a look at this one

0xTim:
@elidydev awesome!

PR #447, opened 2021-08-10
added implementation of async/await handle method for ModelMiddleware 

elidydev:
This pull request implements the async/await `handle` method in ModelMiddleware addressing the issue #446

### DIFF
--- a/Sources/FluentKit/Middleware/ModelMiddleware.swift
+++ b/Sources/FluentKit/Middleware/ModelMiddleware.swift
@@ -54,6 +54,28 @@ extension ModelMiddleware {
         }
     }
     
+    #if compiler(>=5.5)
+    @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
+    public func handle(_ event: ModelEvent, _ model: AnyModel, on db: Database, chainingTo next: AnyModelResponder) async throws {
+        guard let modelType = model as? Model else {
+            return try await next.handle(event, model, on: db).get()
+        }
+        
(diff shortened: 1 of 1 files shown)


mxcl:
Unless I'm missing something, the handle function isn’t called.

0xTim (changes requested):
@mxcl is correct - the async version of handle is nice but we don't call it from anywhere. I think for a first pass we should just call the async versions inside the switch

elidydev:
@0xTim yes, I've seen the note. Thank you for checking in, I have some doubts, so some help would be appreciated.
Maybe I'm missing something, but I don't see a way to call async versions inside the switch of `handle` other than conditionally checking the availability. However, in this case, only the async version will be called on MacOS12, which doesn't seem right.

For this reason, I started looking again into async version of `handle` and make sure it's called. This, however, looks quite invasive so far (for the first issue) and would require async methods in Model+CRUD. If that's okay, I will proceed in this direction.

0xTim:
Yeah this isn't going to be an easy one! 😅 

I think we might need to roll this back and start again. I suggest we add an `AsyncModelMiddleware` protocol that conforms to `ModelMiddleware`. That should solve any chaining issues, ordering issues and where to put the availability check. See https://github.com/vapor/vapor/blob/async-await/Sources/Vapor/Concurrency/AsyncMiddleware.swift as an example.

Are you happy to take that on?

elidydev:
Perfect, thank you for helping out, I'll have a look.

PR #448, opened 2021-08-11
ensure optional enum keys are represented for bulk insert

### LINKED ISSUE #396: Model + CRUD: Order of the elements should not matter when CREATE is called on a collection.
## Description
If we are calling create on a collection of models, depending on the order of collection either an error is thrown or the value is set incorrectly without any error being thrown. For example, the below collection incorrectly set the enum value to `null` and does not throw any error.

```swift
        let workouts = [
            Workout(name: "Walking", duration: 30, distance: 2.5, effort: nil),
            Workout(name: "Running", duration: 30, distance: 4.0, effort: nil),
            Workout(name: "Biking", duration: 30, distance: 5.25, effort: nil),
            Workout(name: "Swimming", duration: 30, distance: 3.5, effort: nil),
            Workout(name: "Yoga", duration: 30, distance: nil, effort: .moderate)
        ]
        return workouts.create(on: database)
    }
```

Here is the debug log for the above code.
```
INSERT INTO "workouts" ("name", "durartionInMinutes", "distanceInKilometers", "id") VALUES ($1, $2, $3, DEFAULT), ($4, $5, $6, DEFAULT), ($7, $8, $9, DEFAULT), ($10, $11, $12, DEFAULT), ($13, $14, NULL, DEFAULT) RETURNING "id" ["Walking", 30.0, 2.5, "Running", 30.0, 4.0, "Biking", 30.0, 5.25, "Swimming", 30.0, 3.5, "Yoga", 30.0]
```

However, if we set the optional enum value in the first item of the collection and do not set it for the rest then it throws the below error. 
``` swift 
        let workouts = [
            Workout(name: "Yoga", duration: 30, distance: nil, effort: .moderate),
            Workout(name: "Walking", duration: 30, distance: 2.5, effort: nil),
            Workout(name: "Running", duration: 30, distance: 4.0, effort: nil),
            Workout(name: "Biking", duration: 30, distance: 5.25, effort: nil),
            Workout(name: "Swimming", duration: 30, distance: 3.5, effort: nil)
        ]
        return workouts.create(on: database)
    }
```

```
NIO-ELT-0-#3 (5): Fatal error: Non-uniform query input: [[id: default, name: "Yoga", effort: moderate, distance_in_kilometers: nil, durartionInMinutes: 30.0], [id: default, durartion_in_minutes: 30.0, name: "Walking", distanceInKilometers: Optional(2.5)], [id: default, name: "Running", durartionInMinutes` 
```

### Model
```swift
final class Workout: Model, Content {
    static let schema = SchemaConstants.workout
    
    @ID(custom: "id", generatedBy: .database)
    var id: Int?
    
    @Field(key: "name")
    var name: String
    
    @Field(key: "durartion_in_minutes")
    var duration: Double
    
    @Field(key: "distance_in_kilometers")
    var distance: Double?
    
    @OptionalEnum(key: "effort")
    var effort: EffortType?
    
    init() {}
    
    init(id: Int? = nil,
         name: String,
         duration: Double,
         distance: Double?,
         effort: EffortType? = nil) {
        self.id = id
        self.name = name
        self.duration = duration
        self.distance = distance
        self.effort = effort
    }
}
```
### Migration
```swift
struct CreateWorkout: Migration {
        return database.schema(SchemaConstants.workout)
            .field("durartion_in_minutes", .double, .required)
            .field("distance_in_kilometers", .double)
            .unique(on: "name")
            .create()
            .map { res in
                return res
        }.flatMap { _ in
            database.enum("effort_type")
                .read()
                .flatMap { efforts in
                database.schema(SchemaConstants.workout)
                    .field("effort", efforts)
                    .update()
            }
        }.transform(to: ())
    }
    
        return database.schema(SchemaConstants.workout).delete()
    }
}
```

### LINKED ISSUE #444: Cannot update value of an @OptionalEnum field to nil
I attempt to update a field defined as an @OptionalEnum to **nil**. The field value does not change. Changing the value to other values of the enum **does** work correctly.

I have also seen this error with other optional fields, i.e. I can update the field to other values, but not to nil

I have shown this to Tim Condon who showed me how to see the error occurring in the debug log, shown below.

### To Reproduce

**FIELD DEFINITION:**	
	@OptionalEnum(key: "playerRequest")
	var playerRequest: PlayerRequestType?

**CODE TO SET OPTIONAL ENUM TO NIL:**
	sportPlayerPivotForPlayerOut.playerStatus = .notPlaying
	sportPlayerPivotForPlayerOut.playerRequestComments = "removed by organiser"
	**sportPlayerPivotForPlayerOut.playerRequest = nil**	//TIM??? !!!	not working
	sportPlayerPivotForPlayerOut.$currentLeague.id = nil
	sportPlayerPivotForPlayerOut.leagueNumber = nil
	let update2 = sportPlayerPivotForPlayerOut.update(on: req.db)
	updates.append(update2)

DEBUG TRACE SHOWING VALUE -NOT- UPDATED:

[ DEBUG ] query update sport_player_pivots filters=[sport_player_pivots[id] = 466EFFB4-8DD7-4812-A66F-29D048E20B49] input=[[playerStatus: notPlaying, dateModified: "2021-07-20T10:38:29Z", leagueNumber: nil, playerRequestComments: "removed by organiser", **playerRequest: remain**, currentLeagueID: nil]] [database-id: psql, request-id: 6BD2882E-DAB7-4985-A1A6-9803B4C59E3D] (FluentKit/Query/Builder/QueryBuilder.swift:297)

1. Add package with configuration '...'
2. Send request with options '...'
3. See error

### Expected behavior

Field value updated to nil

### Environment

Using **postgres**
framework: 4.48.0
toolbox: 18.3.2

* OS version:

11.4

jaredh159:
I'm running into the exact issue described in #396. I can't do batch model inserts with certain combinations of optional fields and enums. It's a major blocker for my app -- as far as I can tell, my only workaround would be to create individual insert requests for each model. But this is a big problem, I have hundreds of records to insert at a time (think something like an app that ingests logs).

I'm not sure honestly how to fix the underlying issue, but at least here's a failing test. I'm hoping that might kickstart someone with more knowledge than me to fix the bug if it's straightforward.

Failing that, if someone could give me a decent description of how to go about the fix myself, I'm definitely willing to try.

Thanks so much!  I'm loving Vapor so far! 🙏 

Fixes #396.
Fixes #444.

### DIFF
--- a/Sources/FluentKit/Enum/OptionalEnumProperty.swift
+++ b/Sources/FluentKit/Enum/OptionalEnumProperty.swift
@@ -79,6 +79,8 @@ extension OptionalEnumProperty: AnyDatabaseProperty {
     public func input(to input: DatabaseInput) {
         if let value = self.value {
             input.set(value.map { .enumCase($0.rawValue) } ?? .null, at: self.field.key)
+        } else {
+            input.set(.null, at: self.field.key)
         }
     }
 
--- a/Sources/FluentKit/Properties/OptionalField.swift
+++ b/Sources/FluentKit/Properties/OptionalField.swift
@@ -62,7 +62,7 @@ extension OptionalFieldProperty: Property {
             if let value = newValue {
                 self.inputValue = value.flatMap { .bind($0) } ?? .null
             } else {
-                self.inputValue = nil
+                self.inputValue = .null
             }
         }
     }
--- a/Tests/FluentKitTests/FluentKitTests.swift
+++ b/Tests/FluentKitTests/FluentKitTests.swift
@@ -539,6 +539,44 @@ final class FluentKitTests: XCTestCase {
             .paginate(pageRequest2)
             .wait())
     }
+
+    func testBulkInsertWithMixedOptionalsDoesntCrash() throws {
+        enum Effort: String, Codable {
+            case moderate
+            case strenuous
+        }
+
+        final class Workout: Model {
+            static let schema = "workouts"
+
+            @ID var id: UUID?
+
+            @Field(key: "name")
+            var name: String
+
+            @OptionalField(key: "distance")
+            var distance: Double?
+
+            @OptionalEnum(key: "effort")
+            var effort: Effort?
+
+            init() {}
+
+            init(id: UUID? = nil, name: String, distance: Double? = nil, effort: Effort? = nil) {
+                self.id = id
+                self.name = name
+                self.distance = distance
+                self.effort = effort
+            }
+        }
+
+        let db = DummyDatabaseForTestSQLSerializer()
+        let workouts = [
+            Workout(name: "Yoga", distance: nil, effort: .moderate),
+            Workout(name: "Running", distance: 3.1, effort: nil),
+        ]
+        XCTAssertNoThrow(try workouts.create(on: db).wait())
+    }
 }
 
 final class User: Model {
--- a/Tests/FluentKitTests/FluentKitTests.swift
+++ b/Tests/FluentKitTests/FluentKitTests.swift
@@ -541,41 +541,26 @@ final class FluentKitTests: XCTestCase {
     }
 
     func testBulkInsertWithMixedOptionalsDoesntCrash() throws {
-        enum Effort: String, Codable {
-            case moderate
-            case strenuous
-        }
-
-        final class Workout: Model {
-            static let schema = "workouts"
-
-            @ID var id: UUID?
-
-            @Field(key: "name")
-            var name: String
-
-            @OptionalField(key: "distance")
-            var distance: Double?
-
-            @OptionalEnum(key: "effort")
-            var effort: Effort?
-
-            init() {}
-
-            init(id: UUID? = nil, name: String, distance: Double? = nil, effort: Effort? = nil) {
-                self.id = id
-                self.name = name
-                self.distance = distance
-                self.effort = effort
-            }
-        }
-
         let db = DummyDatabaseForTestSQLSerializer()
         let workouts = [
             Workout(name: "Yoga", distance: nil, effort: .moderate),
             Workout(name: "Running", distance: 3.1, effort: nil),
         ]
         XCTAssertNoThrow(try workouts.create(on: db).wait())
+    }
+    
+    func testUpdateOptionalEnumToNil() throws {
+        let db = DummyDatabaseForTestSQLSerializer()
+        let workout = Workout(name: "Yoga", distance: nil, effort: .moderate)
+        _ = try workout.save(on: db).wait()
+        workout.effort = nil
+        XCTAssertNil(workout.effort)
+        _ = try workout.update(on: db).wait()
+        XCTAssertEqual(db.sqlSerializers.count, 2)
+        XCTAssertEqual(
+            db.sqlSerializers.last?.sql,
+            #"UPDATE "workouts" SET "effort" = NULL WHERE "workouts"."id" = $1"#
+        )
     }
 }
 
@@ -684,5 +669,34 @@ final class Planet2: Model {
         self.id = id
         self.name = name
         self.moonCount = moonCount
+    }
+}
+
+final class Workout: Model {
+    enum Effort: String, Codable {
+        case moderate
+        case strenuous
+    }
+    
+    static let schema = "workouts"
+
+    @ID var id: UUID?
+
+    @Field(key: "name")
+    var name: String
+
+    @OptionalField(key: "distance")
+    var distance: Double?
+
+    @OptionalEnum(key: "effort")
+    var effort: Effort?
+
+    init() {}
+
+    init(id: UUID? = nil, name: String, distance: Double? = nil, effort: Effort? = nil) {
+        self.id = id
+        self.name = name
+        self.distance = distance
+        self.effort = effort
     }
 }


Tests/FluentKitTests/FluentKitTests.swift
```diff
@@ -539,6 +539,44 @@ final class FluentKitTests: XCTestCase {
...
+                self.effort = effort
+            }
+        }
+
+        let db = DummyDatabaseForTestSQLSerializer()
+        let workouts = [
+            Workout(name: "Yoga", distance: nil, effort: .moderate),
+            Workout(name: "Running", distance: 3.1, effort: nil),
+        ]
+        _ = workouts.create(on: db)
```
0xTim, line 578: This should be waited and wrapped inside an `XCTAssertNoThrow()`

jaredh159: But it doesn't throw, it calls `fatalError()` -- there's no way to catch that in some XCT* func, right?

0xTim: Doing `try workouts.create(on: db).wait()` in tests will work because the tests aren't running on the event loop

jaredh159:
@0xTim Ok, so I've got the test passing (and all the other tests). The source of the problem is pretty simple to understand. The bulk insert functionality takes the field keys from the first model, and then assumes that the field keys will be the same for every subsequent model. When the first model has an optional db enum that happens to be nil, the fluent-kit omits the field altogether.  I traced it to the `create()` function on the Array model extension, where it maps over the models calling `model.collectInput()`, which calls `input(input: DatabaseInput)` on the `Fields` protocol.

OptionalEnum's implementation of this function was setting nothing if the value was nil.  This works fine for single inserts, or inserts where all of the models have a value set (or not set) for the optional enum, but breaks in the case shown in the test case.  The change I made causes OptionalEnum to always set something, which seems to work.

All that said, I'm way above my pay grade here, I can't tell you if this change is the _correct_ way to fix the issue, only that it does fix it, and all the other tests still pass. I could see that this change might cause unnecessary (but harmless) NULL columns to be added to SQL insert statements that didn't really need them, but I'm not sure if that's really even a problem or not.

jaredh159:
@0xTim any chance you (or someone else?) can take a look at this some time fairly soon? Not sure if you saw the commit and comment I made last week, but I think I've got this fixed.  Appreciate your time, thanks!

0xTim:
@jaredh159 this looks great! Could you add a test to see if it fixes #444 as well? I suspect it will

Sources/FluentKit/Properties/OptionalField.swift
```diff
@@ -62,7 +62,7 @@ extension OptionalFieldProperty: Property {
             if let value = newValue {
                 self.inputValue = value.flatMap { .bind($0) } ?? .null
             } else {
-                self.inputValue = nil
+                self.inputValue = .null
```
jaredh159, line 65: This is the fix for #444

Tests/FluentKitTests/FluentKitTests.swift
```diff
@@ -539,6 +539,29 @@ final class FluentKitTests: XCTestCase {
...
+        ]
+        XCTAssertNoThrow(try workouts.create(on: db).wait())
+    }
+    
+    func testUpdateOptionalEnumToNil() throws {
+        let db = DummyDatabaseForTestSQLSerializer()
+        let workout = Workout(name: "Yoga", distance: nil, effort: .moderate)
+        _ = try workout.save(on: db).wait()
+        workout.effort = nil
+        XCTAssertNil(workout.effort)
```
jaredh159, line 557: I was surprised that this _failed_ before the fix was applied.

jaredh159:
@0xTim Added a test for #444, but my initial fix did not also fix that issue. 

Dug around for a while, and I believe I found a fix for #444 as well, included (w/ test) in the latest commit.  Since the fix for that was different, if you'd rather I revert this commit and open another PR with that fix, let me know.

Thanks for your attention to this.

Once again, I'm still a major newcomer to this codebase (and Swift, really), so I'm just trusting that the test suite is good enough to catch if my changes are problematic in some other way.

0xTim:
Nope I think this is good thanks!

@gwynne any thoughts?

Sources/FluentKit/Enum/OptionalEnumProperty.swift
```diff
@@ -79,6 +79,8 @@ extension OptionalEnumProperty: AnyDatabaseProperty {
     public func input(to input: DatabaseInput) {
         if let value = self.value {
             input.set(value.map { .enumCase($0.rawValue) } ?? .null, at: self.field.key)
+        } else {
+            input.set(.null, at: self.field.key)
```
gwynne, line 83: IIRC, this is wrong - the result will be that the enum property's value is _always_ included in `UPDATE` queries, even if it hasn't changed, which breaks various things. I feel like the actual issue is in the setter for `value`, but I can't prove it until I tinker with the code a bit (and I may be wrong). Can you add a unit test that checks that the enum's value is _not_ unconditionally set to `nil` even if there's no corresponding assignment to the property?

jaredh159: @gwynee Thanks for taking a look! Re: the unit test, do you mean something like this?

```swift
    func testUpdateDoesntContainNeedlessNullForNilOptionalEnum() throws {
        let workout = Workout(name: "Yoga", distance: nil, effort: nil)
        _ = try workout.create(on: db).wait()
        workout.name = "Goat Yoga"
        _ = try workout.update(on: db).wait()
        XCTAssertEqual(
            db.sqlSerializers.last?.sql.contains(#""effort" = NULL"#),
            false
        )
    }
```

If so, that test fails.  But I guess I don't see the harm. Technically the `SET "effort" = NULL` is redundant in this case (it already _is_ null), but that's no big deal right?  What scenarios "break" due to this?  And, I guess, why are there no tests for those scenarios?

Is there an optimization somewhere where if you call `model.update(on:)` without anything having been changed, that it's a noop?  I could see this "breaking" that case...  But, I guess I would rather have two gnarly bugs fixed and lose that optimization than keep the optimization and bugs.  That said, I _totally_ am in over my head here, and would love it if you tinkered some and found a more correct solution. Thanks again!

gwynne: Having the redundant SET overwrites data. Consider this sequence:
```swift
SomeModel(stringField: "bar", enumField: .foo)
    .save(on: database)
// Some time later, in another route perhaps:
SomeModel.query(on: database).all().flatMap {
    $0.map { $0.stringField = "baz"; $0.update(on: database) }
}.flatten(on: database.eventLoop)
```
This will set the `enumField` value to `NULL` for every single model in the database. This was obviously not the intended effect 😅

jaredh159: Hmm... I would have thought in your scenario that `enumField` would only be set to `NULL` for any models where it was already `nil`.

gwynne: Unfortunately, no - `AnyDatabaseProperty.input(to:)` is invoked for every property on the model during query generation. For optional properties (and this is one of the reasons the `OptionalField/Parent/Child/EnumProperty` variants exist), the value is a _double_-optional; `.none` means "no value is set for this property", `.some(.none)` means "this property is explicitly set to `nil`", and `.some(.some(...))` is "this property was explicitly given a value". Additionally, calling `DatabaseInput.set(_:at:)` with any value whatsosever guarantees that the resulting query will have the given field key listed, including setting `NULL` for `nil`. The "no value is set" case is handled by just not calling the method for that property at all - this is how the `Optional*Property` types work. The non-optional properties do the same too; they just never pass `nil` as the value when there is one.

gwynne: (It used to be that `FieldProperty` tried to handle both optional and non-optional fields in a single type, but in the end Tanner found that it caused too much havoc, partially due to Swift's somewhat unintuitive behavior around double-optionals.)

jaredh159: @gwynne thanks so much for the in-depth explanation.  I see now that I really truly don't understand the inner workings of this library, and it seems pretty clear that the test suite is not a sufficient safety net for this type of change. 

I don't suppose you'd be willing to take the failing test cases I wrote and work on the correct solution at some point?  I'm very willing to help in any way I can, but after reading your explanation, I don't know if I'm the right person for the job.

The #396 bug is causing me a significant problem in a Vapor app I've just put into production. Because we currently only have a few users (but that number will be growing), as a temporary workaround I'm literally looping over hundreds of records at a time and calling `model.create(on:)` for each one, because I can't do a bulk insert without risking a fatal error. Without this getting fixed, I'm thinking I might need to drop down and write the bulk insert query in raw SQL, which I'm really hoping not to have to do, as I wouldn't feel super confident that I was getting it right for all of the edge cases, and escaping, etc.

jaredh159:
@0xTim @gwynne I'm a bit stumped as to how to proceed with this issue, wondering if you could provide some input.

It seems like there's no one with the time and knowledge of this code base to tackle these two OptionalEnum bugs right now. I think I'm going to resort to writing the raw SQL for my bulk inserts for the time being, but I did have two thoughts about how to proceed:

**First**, I thought (if it sounded agreeable) I might open one or more additional PRs where I just add some more tests around the existing functionality of SQL generation and optional fields, for a few reasons.  One, it seems like the codebase could use more tests. Second, I'm hoping if I write some tests seeking to gain the understanding of how the system is supposed to work, I might eventually work my way to enough knowledge of fluent-kit's inner workings to be able to correctly fix these two issues. If I don't get there, at least we'd have some more tests for future contributors to rely on. Would some PRs adding to the test suite be acceptable, and is there someone who would have the knowledge/bandwidth to review those PRs?

**Second,** would the core team be open to me adding a notice callout on the vapor docs site in the section about `OptionalEnum` (currently just a sentence) stipulating that currently OptionalEnums are significantly hamstrung by a) causing fatal errors during bulk operations, and b) not being able to be nil-ed out?  It seems to me that others should be aware that OptionalEnums are not ready for full-scale production use at the current time.

Thanks for your time 👍

gwynne:
Closing this out, superseded by #469

PR #452, opened 2021-08-20
Use new concurrency verbage

kevinrenskers:
Very similar to https://github.com/vapor/vapor/pull/2672: this just updates the `async-await` branch to use the new name `completeWithTask` rather than the old `completeWithAsync`.


0xTim (approved):
Thanks!

PR #454, opened 2021-09-14
Set method for group fields + tests

JeronimoPaganini:
Since in current version I didn't find how to use similar way to setting values for fields marked like Group, I propose my small improvement. 

Example:

```swift
try Galaxy
    .query(on: self.database)
    .set(\.$size.$km, to: oneLightYearInKm)
    .set(\.$size.$lightYear, to: 1)
    .update()
``` 
where $size is Group fields in a model

```swift
    @Group(key: "area")
    public var size: Size
```

```swift
public final class Size: Fields {
    @Field(key: "km")
    var km: Double
    
    @Field(key: "light_year")
    var lightYear: Double
    
    public init() { }
}
```

If I misunderstood something, please, let me know :)

### DIFF
--- a/Sources/FluentBenchmark/SolarSystem/Galaxy.swift
+++ b/Sources/FluentBenchmark/SolarSystem/Galaxy.swift
@@ -8,6 +8,9 @@ public final class Galaxy: Model {
 
     @Field(key: "name")
     public var name: String
+    
+    @Group(key: "area")
+    public var size: Size
 
     @Children(for: \.$galaxy)
     public var stars: [Star]
@@ -20,6 +23,16 @@ public final class Galaxy: Model {
     }
 }
 
+public final class Size: Fields {
+    @Field(key: "km")
+    var km: Double
+    
+    @Field(key: "light_year")
+    var lightYear: Double
+    
+    public init() { }
+}
+
 public struct GalaxyMigration: Migration {
     public init() {}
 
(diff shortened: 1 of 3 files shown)


gwynne:
@Joannis I need to look at this implementation more closely in any event, but am interested in your input, specifically re: the test failures for Mongo integration where we're apparently hitting the "didn't fetch or set before accessing" codepath?

JeronimoPaganini:
Hi @gwynne , I guess you meant @JeronimoPaganini , 
I'm a bit confused how it happened, since I've reused the same method set, and I've added just +1 signature here:
https://github.com/vapor/fluent-kit/pull/454/files#diff-060bb4b838cade73b63ca17a0ce5a59743bcc45b4ba559f4673bb3a6f79b9726R40

So the plan was just supporting group kodepath (I've re-checked it in my project using Postgres, but didn't check how it works with Mongo). 

Could it be, since mongo-kit doesn't work with group codepaths? (Sorry for the stupid question, I didn't work with mongo-kit yet)

Joannis:
@JeronimoPaganini no, she meant me. I'm maintaining all MongoDB related components.

JeronimoPaganini:
@Joannis, Ah, got it. We have the same 1st letter nickname, and since I've prepared this PR, I thought it's just T9 :)

gwynne:
Closing due to age, please feel free to reopen if this is still an issue.

PR #459, opened 2021-09-30
Bypass Mirror when possible for a 100% speedup in model properties lookup

gwynne:
Uses some very evil underscored runtime function call hackery to accomplish this result. Limited by conditionals to only compiler versions where it is known to work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,10 +42,13 @@ jobs:
         image: mongo:latest
       mongo-b:
         image: mongo:latest
-    container: swift:5.4-focal
+    container: ${{ matrix.container }}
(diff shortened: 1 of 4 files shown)


0xTim (approved):
Overall no objections. We'll need to revisit with each swift version to bump the supported version when we check it still works

gwynne:
@0xTim Worth noting that the `nightly-main` images are already failing the `compiler(<5.6)` check and falling back to the slow path. I'm of a mind to leave it that way since 5.6 is still under active development.

0xTim:
Yeah let's revisit when it gets closer to a 5.6 release. The schedule hasn't been posted yet

PR #461, opened 2021-10-15
Introduce `unique` field constraint

davdroman:
This PR introduces `DatabaseSchema.FieldConstraint.unique`, to allow for uniquing a single field within its definition:

```swift
// Before
db.schema("users")
    .field("email", .string, .required)
    .unique(on: "email")

// After
db.schema("users")
    .field("email", .string, .required, .unique)
```

### DIFF
--- a/Sources/FluentBenchmark/Tests/UniqueTests.swift
+++ b/Sources/FluentBenchmark/Tests/UniqueTests.swift
@@ -13,7 +13,7 @@ extension FluentBenchmarker {
             do {
                 try Foo(bar: "a", baz: 2).save(on: self.database).wait()
                 XCTFail("should have failed")
-            } catch _ as DatabaseError {
+            } catch let error where error is DatabaseError {
                 // pass
             }
         }


davdroman:
Mongo seems to be having issues with this. Any advice?

gwynne:
Pushed a merge from `main` to get a more updated test workflow run.

gwynne:
From what I can see, at the moment the Mongo driver doesn't currently support _any_ of the field-level constraints (`.required`, `.identifier`, `.foreignKey`), and explicitly ignores table-level `.foreignKey` as well. We could work around this by having the `SchemaBuilder` translate the field-level `.unique` into the table version, but I'd prefer to see `FluentMongoDriver` handle this case correctly. @Joannis ?

Joannis:
@gwynne I'm happy introducing `.unique()` in FluentMongoDriver, as that is supported by MongoDB. Just need to know what the expectation is when a single schema has multiple unique constraints. Do I make a compound index, ensuring the _combination_ of those keys is unique? Or do I make sure each field is unique by themselves?

As for `required`, I could add schema validation on collections. But I don't believe most people that use that feature should be using MongoDB.

The other attributed don't exist in MongoDB.

gwynne:
@Joannis A field-level `.unique` constraint over any given `"field"` is exactly equivalent to specifying `.unique(on: "field")` on the table (except under very specific circumstances that Fluent doesn't need to care about). It's legal (at least in SQL) to have multiple table-level unique constraints that specify multiple different sets of columns that make up unique indexes as well - not to mention that `.unique(on: "a", "b")` and `.unique(on: "b", "a")` are two different indexes and check uniqueness differently (shows up immediately when dealing with partial and prefix indexes, not to mention even the flipped ordering can change whether a value is considered unique due to, say, Unicode collation). Oh, and SQL is usually happy to let you go `.unique(on: "a", "b", name: "a").unique(on: "a", "b", name: "b")` and oh look two separate unique constraints that form the same index...

As far as `.required` goes, in SQL that translates to `NOT NULL` (and in turn to not-`Optional` in Swift). If that concept doesn't make sense in MongoDB (especially if it doesn't have an equivalent to `.default` for specifying a value to use when none is given), then ostensibly every `@Field` in a Mongo model would have to be an `@OptionalField` - If that's the case then so be it 🤷‍♀️.

PR #468, opened 2021-11-17
Add extra set method for `DatabaseQuery.Value`

madsodgaard:
This allows the use `DatabaseQuery.Value` when setting a `@Field` property value. For example you can now use SQL:
```swift
query.set(\.$field, to: .sql(raw: "1"))
```

### DIFF
--- a/Sources/FluentBenchmark/Tests/SetTests.swift
+++ b/Sources/FluentBenchmark/Tests/SetTests.swift
@@ -3,6 +3,7 @@ extension FluentBenchmarker {
         try self.testSet_multiple()
         try self.testSet_optional()
         try self.testSet_enum()
+        try self.testSet_queryValue()
     }
     
     private func testSet_multiple() throws {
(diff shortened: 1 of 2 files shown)


0xTim (approved):
LGTM but I'll let @gwynne have the final say

gwynne:
@madsodgaard Adding this overload would make it much easier to violate correctness of field data types, so I'm a little hesitant. Do you have a concrete use case in mind?

madsodgaard:
@gwynne I am using it to have SQL as a value instead of having to drop down to `SQLKit`. Would you prefer having explicit methods for SQL instead? I mean the obvious thing for the user would still be using the correct field data type, and requires them to explicitly state they want to bypass the data type correctness by typing `.`
```swift
Model
  .query(on: database)
  .set(\.$balance, to: .sql(embed: "\(Model.path(for: \.$balance).first!) - \(bind: amount)"))
  .update()
```

gwynne:
@madsodgaard Ideally, I'd prefer to add functionality to FluentKit to express what you're currently using SQL for, rather than encouraging breaking the abstraction. In practice, of course there's only so much that can be supported, and adding an expressive feature to FluentKit isn't usually very easy. (Though, if what you want to do does boil down to "using a binary expression to express an updated column value, as in your example, I do have a pretty clear idea of how to make that work in Fluent with Swift operators - it'd probably end up being as simple as being able to write `.set(\.$balance, to: \.$balance - amount)` directly, similarly to what you can do in `.filter()` and `.join()` clauses.)

gwynne:
Closing this out due to age and lack of response

PR #469, opened 2021-11-18
prevent fatalError when bulk inserting optional enum fields

jaredh159:
This is a hopefully more correct approach to fixing #396, which I first attempted in #448.

The main issue here is that Fluent _omits_ optional NULL values for enums in insert operations. However, when performing a bulk insert, this results in arrays of different lengths being collected if there is at least one model containing `nil` for an optional enum, and one model with a value.  In that case, the app will [trap here](https://github.com/vapor/fluent-kit/blob/main/Sources/FluentSQL/SQLQueryConverter.swift#L119). For example, this seemingly innocuous bit of code currently traps in production:

```swift
let things = [Thing(id: 1, fizzBuzz: nil), Thing(id: 2, fizzBuzz: .fizz)]
try things.create(on: db)
```

My first naive attempt to fix this just made sure that optional enum input values always collected a NULL/nil value. But @gwynne pointed out that this broke some very serious assumptions (which were not covered by any test cases). The tests I added in #455 were an attempt to increase the test coverage around some of these assumptions, and to lay the groundwork for this PR.

In this PR, I am trying to fix it in a more correct (hopefully) way, by introducing the concept of a "collection strategy" -- i.e., a strategy for collecting input.  It's an enum that defaults to the existing behavior of skipping optional enums.  But, in the case of bulk-inserting models, I explicitly set the strategy to `.bulkInsert`, which then causes NULL values to be explicitly set, preventing the problem of non-uniform query input.

I'm not sure about the naming or placement of the new type, but at least would love some input on whether this feels more safe and correct to some of the folks with a deeper knowledge of the library.

### DIFF
--- a/Sources/FluentKit/Enum/EnumProperty.swift
+++ b/Sources/FluentKit/Enum/EnumProperty.swift
@@ -75,7 +75,7 @@ extension EnumProperty: AnyDatabaseProperty {
         self.field.keys
     }
 
-    public func input(to input: DatabaseInput) {
+    public func input(to input: DatabaseInput, strategy: CollectStrategy = .default) {
         if let value = self.value {
             input.set(.enumCase(value.rawValue), at: self.field.key)
         }
--- a/Sources/FluentKit/Enum/OptionalEnumProperty.swift
+++ b/Sources/FluentKit/Enum/OptionalEnumProperty.swift
@@ -76,9 +76,11 @@ extension OptionalEnumProperty: AnyDatabaseProperty {
         self.field.keys
     }
 
-    public func input(to input: DatabaseInput) {
+    public func input(to input: DatabaseInput, strategy: CollectStrategy = .default) {
         if let value = self.value {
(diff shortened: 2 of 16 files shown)


gwynne:
@jaredh159 After looking into this further, this turns out to be two different bugs - one in `OptionalEnumProperty`'s `value` setter (doesn't correctly forward the double-optional), and one in `OptionalFieldProperty`'s `value` getter (incorrectly returns `.none(.none)` instead of `.some(.none)` when the input value is `.null`). I've opened #473 to cover the appropriate fixes - can you test whether this solves your issue?

jaredh159:
@gwynne thanks for looking into this, i'll try to test out your fix tomorrow and get back to you.

0xTim:
Ping @jaredh159 do we need to keep this open or can it be closed?

jaredh159:
Yeah, I think we can close in favor of #469.  I was never able to fully test that patch, but I'm pretty confident if it passes the tests that I wrote (which it does), that it should fix my use case in production.

PR #470, opened 2021-11-20
Fix `AsyncModelMiddleware`

0xTim:
Fixes an issue where creating a middleware that conforms to `AsyncModelMiddleware` would require you to manually implement the `EventLoopFuture`-based `handle(_:_:on:)` which was almost impossible.

This makes `AsyncModelMiddleware` seamless and how it should have been when implemented. Also adds tests to ensure it actually works.

Also fixes a data race and incorrect double-serialization of models when using the `Collection<Model>.create(on:)` CRUD method.

### DIFF
--- a/Sources/FluentKit/Concurrency/AsyncModelMiddleware.swift
+++ b/Sources/FluentKit/Concurrency/AsyncModelMiddleware.swift
@@ -14,24 +14,32 @@ public protocol AsyncModelMiddleware: AnyModelMiddleware {
 
 @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
 extension AsyncModelMiddleware {
-    public func handle(_ event: ModelEvent, _ model: AnyModel, on db: Database, chainingTo next: AnyAsyncModelResponder) async throws {
-        guard let modelType = model as? Model else {
-            try await next.handle(event, model, on: db)
-            return
+    public func handle(_ event: ModelEvent, _ model: AnyModel, on db: Database, chainingTo next: AnyModelResponder) -> EventLoopFuture<Void> {
+        let promise = db.eventLoop.makePromise(of: Void.self)
+        promise.completeWithTask {
+            guard let modelType = model as? Model else {
+                try await next.handle(event, model, on: db).get()
+                return
+            }
+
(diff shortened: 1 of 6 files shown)


0xTim:
@gwynne I think the failures are legitimate TSAN issues as https://github.com/vapor/fluent-kit/blob/main/Sources/FluentKit/Model/Model%2BCRUD.swift#L160 is now accessed across threads... if we wrap it in something like https://github.com/vapor/vapor/blob/main/Tests/VaporTests/Utilities/ThreadSafe.swift that should fix it. Thoughts?

Sources/FluentKit/Concurrency/AsyncModelMiddleware.swift
```diff
@@ -14,24 +14,32 @@ public protocol AsyncModelMiddleware: AnyModelMiddleware {
...
-        case .update:
-            try await self.update(model: modelType, on: db, next: next)
-        case .delete(let force):
-            try await self.delete(model: modelType, force: force, on: db, next: next)
-        case .softDelete:
-            try await self.softDelete(model: modelType, on: db, next: next)
-        case .restore:
-            try await self.restore(model: modelType, on: db, next: next)
+    public func handle(_ event: ModelEvent, _ model: AnyModel, on db: Database, chainingTo next: AnyModelResponder) -> EventLoopFuture<Void> {
+        let promise = db.eventLoop.makePromise(of: Void.self)
```
davdroman, line 18: Slightly shorter version:

```suggestion
        return db.eventLoop.performWithTask { ...
```

gwynne (approved):
Sigh. What a mess, across the board...

Sources/FluentKit/Concurrency/AsyncModelMiddleware.swift
gwynne: This would require bumping the base required version of `async-kit` to ensure the utility is available, which increases the semver-ness of the change.

0xTim: FluentKit doesn't export AsyncKit (🎉🎉🎉 there's at least one!) so we could bump it in the manifest in a patch version. But I'm going to keep it as is as it's one less place to add a dependency and it keeps it in line with the rest of our concurrency stuff

PR #472, opened 2021-11-28
DbView and DbInjectable

AnarchoSystems:
With this feature, you can have the database dynamically populate almost arbitrary models automatically.

Intended use, version 1:

```
try await Moon.query(on: db)
            .join(parent: \.$planet)
            .project(onto: MoonViewModel.self) // needs to conform to Initializable
            .bind(\.$name, to: \MoonViewModel.name)
            .bind(\.planet.$name, to: \MoonViewModel.planetName)
            .all()
```

Intended use, version 2:

```

struct GovernorViewModel : DbInjectable { //can auto-conform to encodable

   typealias ViewBase = Governor

   @FieldBinding(\.$name) var name
   @FieldBinding(\.planet.$name) var planetName 

}

try await Governor.query(on: db).join(parent: \.$planet).bind(to: GovernorViewModel).all()

```


AnarchoSystems:
There are a few source compatibility issues:

- if you extended the ```QueryBuilder``` and made use of the ```all``` method, the compiler might yell at you that you cannot convert the result type of all (now ```View```) into ```Model```. There are two ways to fix this: 1. constrain the extension to the case that View == View.ViewBase 2. just replace ```Model``` with ```View``` wherever appropriate.
- I made a few methods deliberately only available when ```View``` conforms to ```FluentKit.Model```. If you use any of these methods, the compiler may complain whenever it cannot prove this. For existing code, this should mostly apply to extensions of the ```QueryBuilder```, as any method using the ```QueryBuilder``` already had to provide a type conforming to ```FluentKit.Model```.

AnarchoSystems:
Looking forward to your review. Here's what I'm planning to do beyond what I've already done:

- add tests
- add ways to access child and sibling properties with similar ease

AnarchoSystems:
Hmmmm... The test workflow fails for a very non-obvious reason that I do not understand. Apparently, swift compiler <= 5.3 for linux has some problem that later versions don't have. All those jobs complain about something along the lines of:
swift: /home/build-user/swift/lib/AST/GenericSignature.cpp:817: const (anonymous namespace)::RequirementSource *getBestRequirementSource(swift::GenericSignatureBuilder &, ArrayRef<GSBConstraint<swift::ProtocolDecl *> >): Assertion `bestSource && "All sources were self-recursive?"' failed.

1.	Swift version 5.2.5 (swift-5.2.5-RELEASE)
2.	While emitting IR SIL function "@$s9FluentKit19ChildrenEagerLoader33_F9D324A44A4CAE2E97C19B07B627691ALLV3run6models2on7NIOCore15EventLoopFutureCyytGSayxG_AA8Database_ptFySayq_GcfU0_Sbq_XEfU_".
 for expression at [/__w/fluent-kit/fluent-kit/package/Sources/FluentKit/Properties/Children.swift:215:74 - line:222:17] RangeText="{ (child : To) in
                    switch parentKey {
                    case .optional(let optional):
                        return child[keyPath: optional].id == id
                    case .required(let required):
                        return child[keyPath: required].id == id
                    }
                "

I can't seem to find help on google :(

AnarchoSystems:
I replaced the previous implementation such that the change mostly additive and shouldn't break anything. however, this does sacrifice some conveniences of the new API.

in order to make linux with swift 5.2 work, I had to conditionally compile some of the new code. Hence, the ViewBinder will only be available in 5.3 and later or on non-linux platforms.

0xTim:
@AnarchoSystems we've been discussing this internally and at the current time and state of Fluent is not something we can accept. The breaking changes and dropping Swift 5.2 isn't something we can do at this time. However, longer term we do want to be able to support it so @gwynne wants to look at this for Fluent 5

PR #474, opened 2021-12-03
Add support for `.custom(...)` join clauses

finestructure:
We need to perform a `LEFT JOIN` with an additional constraint on the joined table, i.e. in SQL terms we need to execute:

```sql
LEFT JOIN "versions" AS "pre_release_version" 
ON ("packages"."id" = "pre_release_version"."package_id" AND "pre_release_version"."latest" = 'pre_release')
```

This cannot (as far as I can see) be expressed any other way, because moving the additional constraint out of the `ON` clause will effectively turn the query into an `INNER` query: either the ids match or the extra filter will reject any "outer" rows. The `LEFT JOIN` needs to happen "after" the selection, so to speak.

This PR adds an overload exposing the existing `.custom` `QueryBuilder.Join` case such that we can write this as

```swift
            let p = try Package.query(on: app.db)
                .join(Package.self,
                      PreReleaseVersion.self,
                      on: .custom(#"""
                            LEFT JOIN "versions" AS "pre_release_version"
                            ON "packages"."id" = "pre_release_version"."package_id"
                            AND "pre_release_version"."latest" = 'pre_release'
                            """#)
                )
                .all().wait()
```

(See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1072 for some additional details.)

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Join.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Join.swift
@@ -12,6 +12,19 @@ extension QueryBuilder {
         self.join(Foreign.self, filter.foreign, to: Local.self, filter.local , method: method)
     }
 
+    @discardableResult
+    public func join<Foreign, Local>(
+        _ local: Local.Type,
+        _ foreign: Foreign.Type,
+        on join: DatabaseQuery.Join
+    ) -> Self
+        where Foreign: Schema, Local: Schema
+    {
+        self.models.append(Foreign.self)
+        self.query.joins.append(join)
+        return self
+    }
+
     /// This will join a foreign table based on a `@Parent` relation
     ///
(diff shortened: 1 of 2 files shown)


gwynne (approved):
I've toyed with adding this particular overload before, for similar reasons. I've held off before now because it's easy to unknowingly create queries that fail in strange ways (specifically, by referring to models in the join that are not explicitly `.join()`ed elsewhere, due to the additional model(s) not getting added to the query's `models` list), and because I've wanted to just expand Fluent's integrated join capabilities instead. That being said, I think that problem is a sufficiently rare use case versus the number of times I've encountered a need for a custom join that I'm willing to just say I'll make joins nicer in Fluent 5 and just add this overload for now 🙂.

finestructure:
Oh, and thanks a lot for merging this! :)

PR #476, opened 2021-12-09
Allow sorting by `@Group` field properties

davdroman:
Adds missing generic functions to allow `@Group` fields to be sorted


0xTim:
@davdroman could you add a test to confirm this works? Thanks!

0xTim (approved):
Thanks!

PR #478, opened 2022-01-05
Move `DummyDatabase` to `XCTFluent`

stevapple:
`DummyDatabase` is only meant to be used in tests, so `XCTFluent` is where it should be.
`FluentKit` will benefit from dropping dependency on the `NIO` module.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,6 @@ let package = Package(
     ],
     targets: [
         .target(name: "FluentKit", dependencies: [
-            .product(name: "NIO", package: "swift-nio"),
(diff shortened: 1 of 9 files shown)


stevapple:
Although this PR is breaking, I didn’t see anyone using `DummyDatabase` out of `FluentKitTests`, so I believe that’s an acceptable `semver-minor` patch.

gwynne (approved):
I'm all for fixing this. I agree there's no need for it to be semver-major.

0xTim (approved):

PR #479, opened 2022-01-05
Add missing joins for `@OptionalParent` properties

grahamburgsma:
Adds missing functions on `QueryBuilder` for joins using `OptionalParentProperty`.

### Example usage:
```swift 
// Model
class Planet: Model {

    @OptionalParent(key: "star_id")
    var star: Star?
}

// Usage
Planet.query(on: db)
    .join(from: Planet.self, parent: \.$star)
    // OR
    .join(parent: \.$star)
```


0xTim:
@grahamburgsma Thanks for this! Could you add a test to ensure it works/remains working?

0xTim (approved):
Thanks!

PR #482, opened 2022-02-07
Add `hasNext` and `hasPrevious` properties

noahpistilli:
This PR adds the properties `hasNext` and `hasPrevious` which are similar to the flask-sqlalchemy properties `has_next` and `has_prev`.


0xTim:
Can you add some tests to this? Especially to check the boundaries/edge cases. Like count of 0, trying to pass in a page of -1 etc

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
```diff
@@ -80,6 +86,20 @@ public struct PageMetadata: Codable {
...
         self.per = per
         self.total = total
+        
+        // Find out how many items are left in the pagination object
+        let remainingItems = total - (per * page)
+        if remainingItems <= 0 {
+            self.hasNext = false
+        } else {
+            self.hasNext = true
+        }
```
lewisgodowski, line 96: How do we feel about one-liners like below?

```suggestion
        self.hasNext = (total - (per * page)) <= 0
        self.hasPrevious = page > 1
```

0xTim: Personally I'd choose the existing code, it's a bit easier to read at a glance. But I don't have a strong opinion on it

noahpistilli:
> Can you add some tests to this? Especially to check the boundaries/edge cases. Like count of 0, trying to pass in a page of -1 etc

Should I add bounds to `PageMetadata.page` and `PageMetadata.per` then? Right now it will gladly accept a negative page number a value of `0`

0xTim:
> > Can you add some tests to this? Especially to check the boundaries/edge cases. Like count of 0, trying to pass in a page of -1 etc
> 
> Should I add bounds to `PageMetadata.page` and `PageMetadata.per` then? Right now it will gladly accept a negative page number and a value of `0`

Yes we definitely need tests and count checking for this

gwynne:
Unfortunately, because `PageMetadata` is both `public` and `Codable`, this is a breaking change, and at this point I'm no longer accepting new features for Fluent 4. That being said, this functionality will be available in Fluent 5.

PR #484, opened 2022-02-20
Improve usefulness of fatalError() messages in Fluent

gwynne:
Fluent's implementation includes a number of uses of `fatalError()` in cases of API misuse. However, the stack backtraces for fatal errors have become less and less reliable as a means of locating the source of the error, especially in cases like Fluent's where the error itself is in completely generic code. To make matters worse, most of the error messages are extremely unhelpful - it's completely impossible to determine, for example, just _which_ child relation on what original model something accessed without loading it first. And so on.

Therefore, as of now, all of the fatal error messages try to make sure they include some kind of description of what specific access or data was responsible for the crash. Hopefully this will make it easier to debug failures, especially when these code paths get hit in production systems.

### DIFF
--- a/Sources/FluentKit/Concurrency/Database+Concurrency.swift
+++ b/Sources/FluentKit/Concurrency/Database+Concurrency.swift
@@ -3,7 +3,7 @@ import NIOCore
 
 @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
 public extension Database {
-    func transaction<T>(_ closure: @escaping (Database) async throws -> T) async throws -> T {
+    func transaction<T>(_ closure: @Sendable @escaping (Database) async throws -> T) async throws -> T {
         try await self.transaction { db -> EventLoopFuture<T> in
             let promise = self.eventLoop.makePromise(of: T.self)
             promise.completeWithTask{ try await closure(db) }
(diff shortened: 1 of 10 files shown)


Sources/FluentKit/Concurrency/Database+Concurrency.swift
```diff
@@ -3,15 +3,15 @@ import NIOCore
 
 @available(macOS 12, iOS 15, watchOS 8, tvOS 15, *)
 public extension Database {
-    func transaction<T>(_ closure: @escaping (Database) async throws -> T) async throws -> T {
+    func transaction<T>(_ closure: @Sendable @escaping (Database) async throws -> T) async throws -> T {
```
tanner0101, line 6: what's `@Sendable`? o_O

tanner0101 (approved):

Sources/FluentKit/Concurrency/Database+Concurrency.swift
0xTim: Oh Tanner....

🤣

0xTim: In all seriousness this shouldn't be API breaking but does it add any extra warnings?

0xTim (approved):
LGTM

jdmcd (approved):
LGTM!

Sources/FluentKit/Concurrency/Database+Concurrency.swift
gwynne: @0xTim It actually corrected two of the four in that file. AFAIK it doesn't add any that wouldn't have showed up anyway for callers.

PR #485, opened 2022-02-21
Adds a page count to `PageMetadata`

fwgreen:
Adds the page count as a computed property to `PageMetadata` so it doesn't need to be manually calculated based off the total number of items and items per page.

Fixes https://github.com/vapor/fluent-kit/issues/481

PR #486, opened 2022-02-21
Add tests for page count

fwgreen:


0xTim:
@fwgreen can you combine this into #485 so the PR has the new feature and the tests? Thanks!

PR #487, opened 2022-02-22
Make it easier to save and restore models with relations outside the database

gwynne:
When trying to cache database lookup results in a faster storage, such as Redis, it is often difficult to store models, as the `Children`, `OptionalChild`, and `Siblings` property wrappers do not round-trip correctly through `Codable`. This doesn't solve the issue, but it does enable users to work around it by manually restoring the necessary ID value, which is not currently possible except by faking a database output (and doing so in turn incurs much of the overhead of Fluent decoding which is otherwise bypassed by cache storage).

### DIFF
--- a/Sources/FluentKit/Properties/Children.swift
+++ b/Sources/FluentKit/Properties/Children.swift
@@ -44,7 +44,8 @@ public final class ChildrenProperty<From, To>
     }
     
     public var fromId: From.IDValue? {
-        return self.idValue
+        get { return self.idValue }
+        set { self.idValue = newValue }
     }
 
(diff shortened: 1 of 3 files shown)


jdmcd (approved):

gwynne:
Additional note - it is not currently possible to just correctly implement the `Codable` conformance, as the format of the encoded output would have to change, breaking public API.

Joannis (approved):
I don't like that we need to resort to solutions like this.. It shows that the next iteration of Fluent has a lot of internal mistakes to learn from. The usecase is very valid, and the solution matching. Approved.

0xTim (approved):

PR #488, opened 2022-03-01
Upsert Support

grahamburgsma:
Adds basic support for upsert/conflict strategy. I'm looking for feedback on the API.

Addresses #50.

### Query
```swift
try Thing.query(on: db)
    .set(\.$name, to: "First")
    .create(
        onConflict: \.$name,
        strategy: .update {
            $0.set(\.$name, to: "Last")
        }
    )
```

### Model
```swift
try Thing(name: "First")
  .create(
    onConflict: \.$name,
    .update {
        $0.set(\.$name, to: "Last")
    },
    on: db
)
```

When used on a Model, the conflict fields are returned by the query and saved back to the model. So if there is a conflict, the model will have all the correct values.

### DIFF
--- a/Sources/FluentKit/Model/Model+CRUD.swift
+++ b/Sources/FluentKit/Model/Model+CRUD.swift
@@ -17,6 +17,17 @@ extension Model {
         create(onConflict: Self.path(for: field), strategy, on: database)
     }
 
+    public func create<Field1: QueryableProperty, Field2: QueryableProperty>(
+        onConflict field1: KeyPath<Self, Field1>,
+        _ field2: KeyPath<Self, Field2>,
+        _ strategy: QueryBuilder<Self>.ConflictStrategy,
+        on database: Database
+    ) -> EventLoopFuture<Void>
+    where Field1.Model == Self, Field2.Model == Self
+    {
+        create(onConflict: [Self.path(for: field1)[0], Self.path(for: field2)[0]], strategy, on: database)
+    }
+
     public func create(onConflict fields: [FieldKey], _ strategy: QueryBuilder<Self>.ConflictStrategy, on database: Database) -> EventLoopFuture<Void> {
         return database.configuration.middleware.chainingTo(Self.self) { event, model, db in
             let conflictStrategy = DatabaseQuery.ConflictResolutionStrategy(fields: fields, strategy: strategy)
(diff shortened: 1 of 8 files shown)


wibed:
is this only for the mongo driver?

grahamburgsma:
> is this only for the mongo driver?

@wibed I've been testing with Postgres so far. It should work for all that Fluent/SQLKit supports. Each driver needs changes to enable it though, that's why it looks like all the tests are failing.

josephxanderson:
Any updates on this? Or #50?

grahamburgsma:
> Any updates on this?

@josephxanderson I never got any feedback on it or much interest in it, so I've kinda abandoned it. I imagine Fluent 5 will address this feature.

0xTim:
cc @gwynne

gwynne:
@grahamburgsma I must apologize; I quite honestly forgot to return to this 😰. While the PR is obviously a bit out of date at this point, it was very well written, and I have no significant complaints about the proffered API. If you're willing to bring the PR up to date with `main`, I'd gladly accept it now.

The only minor concern I have with the API as presented is there doesn't appear to be any way to get the effect of SQLKit's `SQLConflictUpdateBuilder.set(excludedValueOf:)`; I would say it could be left to add later, but Fluent's design has a way of making it difficult to handle features in such an additive fashion without bending or breaking semver rules 🤦‍♀️. I definitely don't consider it a requirement for merging this, but if you have any thoughts on how to provide the additional functionality at the Fluent layer (I admittedly don't, at least not offhand 😅), I'd say go for it!

(Incidentally, you're correct that Fluent 5 will have support for this built in, but just the same, Fluent 4 can still very much benefit from having this available.)

grahamburgsma:
Sorry for the late reply, thanks for looking it over @gwynne! 

Here's where I'm at: I did update it with main and made some changes to get it back into a working state and have it working with Postgres. However when I tested against SQLite my changes had broken other things. I tried getting both to work but fixing one would break the other and still haven't tested the other DBs. Seems there are just too many differences between the inner workings of the different DB drivers.

If anyone is interested in taking over this work feel free to, I'll push my latest changes. For now, I don't have time to invest into this so I will close this PR. We'll have to put all our chips on Gwynne and Fluent 5 for this one 😆

Issue #491, opened 2022-03-10
MySQL CONSTRAINT/KEY logic result in error when trying to delete a constraint


kevinzhow:
https://github.com/vapor/fluent-kit/blob/9d47c328bf83999968c12a3bc94ead1d706ad4a9/Sources/FluentSQL/SQLSchemaConverter.swift#L280

This CONSTRAINT/KEY logic result in this issue https://github.com/vapor/fluent/issues/722 which will made `deleteConstraint` didn't work

Wondering why we have this logic and do we have any unit test on this?

gwynne:
This was handled more comprehensively in #522

PR #492, opened 2022-03-10
Fix mysql SQLDropConstraint CONSTRAINT serialize

kevinzhow:
fix https://github.com/vapor/fluent/issues/722 and https://github.com/vapor/fluent-kit/issues/491

### DIFF
--- a/Sources/FluentSQL/SQLSchemaConverter.swift
+++ b/Sources/FluentSQL/SQLSchemaConverter.swift
@@ -277,11 +277,7 @@ public struct SQLDropConstraint: SQLExpression {
     }
 
     public func serialize(to serializer: inout SQLSerializer) {
-        if serializer.dialect.name == "mysql" {
-            serializer.write("KEY ")
-        } else {
-            serializer.write("CONSTRAINT ")
-        }
+        serializer.write("CONSTRAINT ")
         let normalizedName = serializer.dialect.normalizeSQLConstraint(identifier: name)
         normalizedName.serialize(to: &serializer)
     }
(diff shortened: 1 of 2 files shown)


0xTim:
This looks OK to me but I don't know enough about MySQL to say what is the correct approach. I'll leave it to Gwynne. It may be that different versions require different dialects

kevinzhow:
> This looks OK to me but I don't know enough about MySQL to say what is the correct approach. I'll leave it to Gwynne. It may be that different versions require different dialects

you were right, the old work around is for mysql 5.7, but....
after I testing the old work around it on mysql 5.7, it also result in issue.

according to [MySQL Reference](https://dev.mysql.com/doc/refman/5.7/en/create-table-foreign-keys.html)

```
Dropping Foreign Key Constraints

You can drop a foreign key constraint using the following [ALTER TABLE](https://dev.mysql.com/doc/refman/5.7/en/alter-table.html) syntax:


ALTER TABLE tbl_name DROP FOREIGN KEY fk_symbol;
```

When to MySQL 5.7, it should be `FOREIGN KEY`
MySQL 8.0 and later, it should be `CONSTRAINT`

bottlehall:
Is there any chance of fixing this? My original issue  [722](https://github.com/vapor/fluent/issues/722 remains and it makes unit-testing very tedious as you can't revert the database automatically.

gwynne:
@bottlehall #522 should have solved this issue. Are you still running into it?

bottlehall:
Hi Gwynne

\Yes, I’ve just created a minimal migration/reversion and it still happens. It fails with a MySQL error:

testIndex(): failed: caught error: "MySQL error: Server error: Cannot drop column 'defaultUnitId': needed in a foreign key constraint '864a4b794b34ca99a50c05a6504fa9f21b839e9a’"

This is the field that was added in the migration and the constraint is created during the migration. 

    struct UpdateUserTable2: AsyncMigration {
        func prepare(on database: Database) async throws {
            try await database.schema("User")
                .field("defaultUnitId", .uuid, .references("Unit", "id"))
                .update()
        }

        func revert(on database: Database) async throws {
            try await database.schema("User")
                .deleteField("defaultUnitId")
                .update()
        }
    }

I have checked that the same error arises if I use the alternative method of creating the foreign key:

.foreignKey("createdByUserId", references: "User", "id", name: "created_user_id”)

And reverting using:

.deleteConstraint(name: "created_user_id")

Before trying to delete the field.

Best wishes

Nick

gwynne:
Support for the missing functionality and information regarding its use has been added in #575

PR #498, opened 2022-04-08
Further improve the workaround for Mirror's slowness

gwynne:
Refactors the workaround into its own `IteratorProtocol` and `Sequence` implementation, and starts using the workaround for `AnyModel.anyID` for good measure (this was not happening before).

Also removes unneeded `@testable` imports from the tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,6 +53,8 @@ jobs:
         swiftver:
           - swift:5.2
           - swift:5.5
+          - swift:5.6
+          - swiftlang/swift:nightly-main
         swiftos:
           - focal
         dependent:
@@ -66,11 +68,11 @@ jobs:
         run: apt-get -q update && apt-get -q install -y libsqlite3-dev
         if: ${{ contains(matrix.dependent, 'sqlite') }}
       - name: Check out package
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           path: package
       - name: Check out dependent
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           repository: vapor/${{ matrix.dependent }}
           path: dependent
--- a/Sources/FluentKit/Model/AnyModel.swift
+++ b/Sources/FluentKit/Model/AnyModel.swift
@@ -34,10 +34,15 @@ extension AnyModel {
     }
 
     var anyID: AnyID {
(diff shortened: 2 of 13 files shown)


.github/workflows/test.yml
```diff
@@ -53,6 +53,8 @@ jobs:
         swiftver:
           - swift:5.2
           - swift:5.5
+          - swift:5.6
+          - swiftlang/swift:nightly-main
```
0xTim, line 57: I don't think this will work, there's a regression with the compiler and Vapor

0xTim (approved):
LGTM I think!

Sources/FluentKit/Model/AnyModel.swift
```diff
@@ -34,10 +34,15 @@ extension AnyModel {
     }
 
     var anyID: AnyID {
-        guard let id = Mirror(reflecting: self).descendant("_id") as? AnyID else {
-            fatalError("id property must be declared using @ID")
+        for (nameC, child) in _FastChildSequence(subject: self) {
```
Joannis, line 37: Can we get some docs? 😇

Joannis: There's a few constants I'd love to understand

Joannis:
What are the performance differences? I know it can only be good.

.github/workflows/test.yml
gwynne: Apparently Fluent by itself doesn't hit whatever's borked 🤷‍♀️

Sources/FluentKit/Model/AnyModel.swift
gwynne: Okay, yeah, that's fair, I should've at least added the comments for the ASCII codes 😅

gwynne:
> What are the performance differences? I know it can only be good.

The original workaround, before this PR, halved the runtime of the `properties` and `labeledProperties` methods. It does even better for `anyID`, something on the order of a 5x improvement - the ID property tends to be at or near the beginning of most models (and the compiler is usually cooperative about laying out the field metadata in the order they appear in the source), so the early exit from the iteration loop on match yields big wins. I don't know yet how that translates to Fluent performance in general; `anyID` is used in a lot of places, but it's not the real bottleneck (the real one is the `properties` method).

PR #501, opened 2022-04-25
Make accessing `model.id` safe

### LINKED ISSUE #495: Shouldn't use `model.id!` but `guard let id = model.id`
### Describe the bug

The codebase uses lots of `model.id!` e.g. https://github.com/vapor/fluent-kit/blob/main/Sources/FluentKit/Model/Model%2BCRUD.swift#L74, and if a model was newly created using a custom `init` without generating the `id` yet (e.g. on iOS client), accessing `model.id!` causes instant crash. Instead, we should `guard` and throw an exception.

### To Reproduce

1. Add a model and a custom `init`, and don't manually assign a value to `id`
2. Call `model.delete()`

### Expected behavior

An exception should be thrown as `delete` etc are throwing functions

### Environment

I'm using `fluent-sqlite-driver` of `main` branch via SPM in Xcode on iOS 15.4, which depends on `FluentKit`

* Vapor Framework version: NA
* Vapor Toolbox version: NA
* OS version: macOS 12.3

ahmdyasser:
This removes force unwraps when accessing a model's ID in the CRUD operations and changes them to throw an error instead to make the code safer.

Resolves #495

### DIFF
--- a/Sources/FluentKit/Model/Model+CRUD.swift
+++ b/Sources/FluentKit/Model/Model+CRUD.swift
@@ -9,7 +9,7 @@ extension Model {
 
     public func create(on database: Database) -> EventLoopFuture<Void> {
         return database.configuration.middleware.chainingTo(Self.self) { event, model, db in
-            model.handle(event, on: db)
+            try model.handle(event, on: db)
         }.handle(.create, self, on: database)
     }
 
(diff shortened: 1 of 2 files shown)


0xTim (approved):
Thanks!

0xTim:
Original comment:
- replacing model.id! with guard let id = model.id
- marking the containing function as throwing function
- throw error if accessing id before initializing it

hyouuu:
Awesome!

PR #509, opened 2022-05-29
Initial support for composite primary keys

gwynne:
@tanner0101 _Look upon my works, ye mighty, and despair..._ 🤣

### DIFF
--- a/Sources/FluentKit/Properties/Property.swift
+++ b/Sources/FluentKit/Properties/Property.swift
@@ -32,10 +32,17 @@ public protocol AnyCodableProperty: AnyProperty {
 
 public protocol AnyQueryableProperty: AnyProperty {
     var path: [FieldKey] { get }
+    func queryableValue() -> DatabaseQuery.Value?
 }
 
 public protocol QueryableProperty: AnyQueryableProperty, Property {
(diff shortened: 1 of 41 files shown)


0xTim (commented):
Initial work LGTM

Sources/FluentKit/Properties/Property.swift
```diff
@@ -32,14 +32,44 @@ public protocol AnyCodableProperty: AnyProperty {
 
 public protocol AnyQueryableProperty: AnyProperty {
     var path: [FieldKey] { get }
+    func queryableValue() -> DatabaseQuery.Value?
```
0xTim, line 35: Is this technically a breaking change for anyone who creates their own properties?

0xTim (approved):
One query just to make sure we go into this with eyes open. I've tested this with FluentPostGIS as well and it's all good

Sources/FluentKit/Properties/Property.swift
gwynne: No, because a default implementation is provided (and it's not legal to conform to `AnyQueryableProperty` without conforming to `QueryableProperty`).

PR #514, opened 2022-06-18
Add missing info to missingParent error, fix ID compare for composite IDs

gwynne:
- The `missingParent` case of `FluentError` now contains detailed information about which model was missing as part of the error itself. (Yes, this is technically a public API-breaking change, but I'm invoking the "enums are a messy edge case for semver" exception because it's a real problem that this info has been missing from the API.)
- `.filter(\.$id == ...)` and `.filter(\.$id != ...)` now work when the queried model is using `@CompositeID`.


jdmcd (approved):
LGTM - thank you

PR #519, opened 2022-06-25
Additional API for pagination to access individual page

nashysolutions:
Adds the ability to get a single page when paginating a request:

```swift
let page = try await Model.query(on: req.db).page(withIndex: 2, size: 10)
```


0xTim (approved):
Looks useful thanks!

0xTim (changes requested):
Could you add an async wrapper for the new function to QueryBuilder+Concurrency.swift? Thanks!

0xTim:
Oh we'll also need tests for the new behaviour, especially around the boundaries like requesting page `-1`  or page 11 when there are only 10 pages etc

PR #521, opened 2022-06-30
Formatted boolean properties

gwynne:
Adds two new property types to Fluent: `@Boolean` and `@OptionalBoolean`. These properties always have `Bool` values, and can be configured with a storage format, similarly to `@Timestamp`. For example, the `.trueFalse` format stores the strings `"true"` and `"false"` in the database, and parses those strings on load. If no format is specified, the database's default native format for `Bool` is used, equivalently to using `@Field` or `@OptionalField`. An appropriate column data type must be specified in the migration for the model.

### DIFF
--- a/Sources/FluentKit/Properties/Boolean.swift
+++ b/Sources/FluentKit/Properties/Boolean.swift
@@ -3,6 +3,45 @@ extension Fields {
         where Format: BooleanPropertyFormat
 }
 
+/// A Fluent model property which represents a boolean (true/false) value.
+///
+/// By default, `Bool` properties are stored in a database using the storage format
+/// defined by the database driver, which corresponds to using the `.bool` data type
+/// on the appropriate field in a migration. This property wrapper allows specifying
+/// an alternative storage format - such the strings "true" and "false" - which is
+/// automatically translated to and from a Swift `Bool` when loading and saving the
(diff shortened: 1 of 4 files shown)


jdmcd (approved):

Sources/FluentKit/Properties/Boolean.swift
```diff
@@ -0,0 +1,82 @@
+extension Fields {
+    public typealias Boolean<Format> = BooleanProperty<Self, Format>
+        where Format: BooleanPropertyFormat
+}
+
+@propertyWrapper
```
0xTim, line 6: Would be good to have some API docs as we expand our our reference docs

Sources/FluentKit/Properties/OptionalBoolean.swift
```diff
@@ -0,0 +1,98 @@
+extension Fields {
+    public typealias OptionalBoolean<Format> = OptionalBooleanProperty<Self, Format>
+        where Format: BooleanPropertyFormat
+}
+
+@propertyWrapper
+public final class OptionalBooleanProperty<Model, Format>
```
0xTim, line 7: Something something API docs

0xTim (approved):
LGTM - would be good to add some API docs but otherwise all good!

PR #523, opened 2022-07-11
Allow updating deleted models with `Model.update(on:)`

grahamburgsma:
Previously could not update a model that was deleted using `Model.update(on:)` which seem unintuitive.

### DIFF
--- a/Sources/FluentKit/Model/Model+CRUD.swift
+++ b/Sources/FluentKit/Model/Model+CRUD.swift
@@ -58,6 +58,7 @@ extension Model {
         let input = self.collectInput()
         guard let id = self.id else { throw FluentError.idRequired }
         return Self.query(on: database)
+            .withDeleted()
             .filter(id: id)
             .set(input)
             .update()


0xTim:
@grahamburgsma looks like this greats all the unit tests because of the expected generated SQL. That will need to be fixed, but waiting to hear back from @gwynne as to whether this should be supported or there's some context we're missing. (I think this is fine)

grahamburgsma:
> That will need to be fixed, but waiting to hear back from @gwynne

For sure I'll fix those if it otherwise looks good from @gwynne.  It is a change in behaviour so I wasn't sure if this would be accepted.

gwynne:
In theory there's nothing wrong with this, but in practice it represents a significant change in the behavior of a public API; I'm not really comfortable with it when there's an existing workaround (in the form `SomeModel.query(on: database).withDeleted().set(\.$foo, to: bar).update()`).

grahamburgsma:
> In theory there's nothing wrong with this, but in practice it represents a significant change in the behavior of a public API; I'm not really comfortable with it when there's an existing workaround.

Yes I was hesitant to even make this PR as people could potentially be relying on the current behaviour, but wanted to see what you thought of it. Thanks for the review!

PR #525, opened 2022-07-21
SchemaBuilder functions with @discardableResult

dmonagle:
Added @discardableResult to all functions in SchemaBuilder that modify Self and return Self. This brings it in line with QueryBuilder and also allows more complicated migrations that use conditional logic (ie not chained) to work without having to use `let _ =` to avoid `Result of call to 'xxx' is unused` warnings.


gwynne (approved):
LGTM! I'm actually amazed no one ever noticed this or opened an issue about it before now.

PR #526, opened 2022-07-25
Log start and finish of migration prepares and reverts

gwynne:
Also adds some documentation comments to the confusion that is `Property.swift`.

Shoutout to @dmonagle for being the first to point out to me that this logging was missing!

### DIFF
--- a/Sources/FluentKit/Migration/Migrator.swift
+++ b/Sources/FluentKit/Migration/Migrator.swift
@@ -246,13 +246,17 @@ private final class DatabaseMigrator {
     // MARK: Private
 
     private func prepare(_ migration: Migration, batch: Int) -> EventLoopFuture<Void> {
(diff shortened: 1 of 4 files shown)


Sources/FluentKit/Migration/Migrator.swift
```diff
@@ -246,13 +246,17 @@ private final class DatabaseMigrator {
     // MARK: Private
 
     private func prepare(_ migration: Migration, batch: Int) -> EventLoopFuture<Void> {
+        self.database.logger.info("[Migrator] Starting prepare of migration '\(migration.name)'")
```
0xTim, line 249: Actually I know I said yes to this but let's move the migration name into the logger metadata instead

0xTim (changes requested):

0xTim (approved):
LGTM

PR #529, opened 2022-08-10
Performance and behavior improvements for encoding/decoding Fluent relations

gwynne:
This PR addresses the following items:

- The performance of eager loading for `@Parent` and `@OptionalParent` relations should now scale more strictly linearly rather than exhibiting eventually quadratic behavior. This only affects the last step of eager loading, and will probably only be noticeable in queries which eager load a very large number of small models.
- The information contained by a `FluentError.missingParent()` error is now more complete and consistent.
- The default `Codable` conformance provided to any type conforming to `Fields` (which includes all `Model`s) should now be slightly but noticeably more performant; a number of unnecessary extra intermediate steps and repeated actions have been removed.
- Errors thrown from the default `Codable` conformance on `Fields` now retain the full coding path of the entire coding operation, rather than only the failing key.
- Setting an `@OptionalParent` property's `value` to `nil` (or more precisely, `.some(.none)`) no longer results in the property failing to encode its `id`.

There is a new protocol requirement on `AnyCodableProperty` - while it is not expected that external users need to be concerned with said requirement, it does technically make this a `semver-minor` update (see the comments on `AnyCodableProperty.skipPropertyEncoding` for details of the new property).


Sources/FluentKit/Properties/CompositeID.swift
```diff
@@ -74,11 +74,13 @@ extension CompositeIDProperty: AnyDatabaseProperty {
 
 extension CompositeIDProperty: AnyCodableProperty {
     public func encode(to encoder: Encoder) throws {
-        try self.value!.encode(to: encoder)
+        var container = encoder.singleValueContainer()
+        try container.encode(self.value!)
```
0xTim, line 78: I know this is probably safe but should we try a guard and throw an error in the case it should never hit?

Sources/FluentKit/Properties/OptionalParent.swift
```diff
@@ -151,25 +151,25 @@ extension OptionalParentProperty: EagerLoadable {
...
-                }.first)
+        var sets = Dictionary(grouping: models, by: { $0[keyPath: self.relationKey].id })
+        let nilParentModels = sets.removeValue(forKey: nil) ?? []
+
+        return To.query(on: database).filter(\._$id ~~ Set(sets.keys.compactMap { $0 })).all().flatMapThrowing {
+            let parents = Dictionary(uniqueKeysWithValues: $0.map { ($0.id!, $0) })
+
+            for (parentId, models) in sets {
+                guard let parent = parents[parentId!] else {
+                    database.logger.debug("No parent '\(To.self)' with id '\(parentId!)' was found in eager-load results.")
```
0xTim, line 167: Parent ID should be metadata instead if embedded in the String

Sources/FluentKit/Properties/Parent.swift
```diff
@@ -146,33 +146,22 @@ extension ParentProperty: EagerLoadable {
...
-                        to: "\(To.self)",
-                        key: From.path(for: self.relationKey.appending(path: \.$id)).first!.description,
-                        id: "\(model[keyPath: self.relationKey].id)"
-                    )
+        return To.query(on: database).filter(\._$id ~~ Set(sets.keys)).all().flatMapThrowing {
+            let parents = Dictionary(uniqueKeysWithValues: $0.map { ($0.id!, $0) })
+
+            for (parentId, models) in sets {
+                guard let parent = parents[parentId] else {
+                    database.logger.debug("No parent '\(To.self)' with id '\(parentId)' was found in eager-load results.")
```
0xTim, line 161: As above

0xTim (approved):
LGTM - a few nits but feel free to take them or leave them

Sources/FluentKit/Properties/CompositeID.swift
gwynne: A `CompositeID` property with no value set will have already crashed out in 20 other places before it gets anywhere near here; the only reason the value is optional is the definition of `id` on `Model` absolutely requires it (despite the fact that it's not a supported scenario for composite ID models 🤦‍♀️).

Sources/FluentKit/Properties/OptionalParent.swift
gwynne: Honestly, I only even added this log line for @jdmcd's benefit, and later mooted the point by making the call to slightly break API by throwing a useful (**GASP!!!**) error; it probably can just be removed.

PR #534, opened 2022-09-28
fix: detach not invoke delete lifecycle

AFutureD:
This PR is aiming to fix the issue #533 and [vapor/fluent#747](https://github.com/vapor/fluent/issues/747#issuecomment-1259817436_).

The `SiblingsProperty.detach(_:on:)` miss to invoke delete lifecycle, while `SiblingsProperty.attach(_:on:)` does.

The last line `.delete()` just simply build an sql and runs it, which cause some issues.

```Swift
    public func detach(_ tos: [To], on database: Database) -> EventLoopFuture<Void> {
        guard let fromID = self.idValue else {...}
        let toIDs = tos.map {...}

        return Through.query(on: database)
            .filter(self.from.appending(path: \.$id) == fromID)
            .filter(self.to.appending(path: \.$id) ~~ toIDs)
            .delete()
    }
```

### DIFF
--- a/Sources/FluentKit/Properties/Siblings.swift
+++ b/Sources/FluentKit/Properties/Siblings.swift
@@ -203,7 +203,7 @@ public final class SiblingsProperty<From, To, Through>
         return Through.query(on: database)
             .filter(self.from.appending(path: \.$id) == fromID)
             .filter(self.to.appending(path: \.$id) ~~ toIDs)
-            .delete()
+            .all().flatMapEach(on: database.eventLoop) { $0.delete(on: database) }
     }
 
     /// Detach a single model by deleting the pivot.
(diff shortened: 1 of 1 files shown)


0xTim (commented):
I think we're going to have to be a bit cleverer here and manually invoke the lifecycle or similar because at the moment we can't accept the patch as is as it turns a functions that performs one SQL query into a function that performs as many queries as there are siblings which would be a big performance hit

0xTim:
cc @gwynne

AFutureD:
Agreed. This PR does have some potential performance issues. But, it might be a little difficult to figure out the cleaver solution. Waiting for your idea.

AFutureD:
If we could get all models first, and then invoke lifecycle manually.
After that, we delete as before?

Pseudocode may as follows:
```
let query = Through.query(on: database)
            .filter(self.from.appending(path: \.$id) == fromID)
            .filter(self.to.appending(path: \.$id) ~~ toIDs)

try await query.all().map { $0.invoke() }

try await query.delete()
```

0xTim:
Unfortunately that still loads potentially a very large number of pivots in to memory which can be slow. I'm going to defer to @gwynne when she gets a chance to look at it as it's her codebase and she's the expert on all things Fluent

PR #539, opened 2022-11-04
Allow use of create and delete for arrays of models that have composite IDs.

gwynne:
We no longer crash in a precondition failure when attempting to invoke Fluent's `.create()` and `.delete()` extensions to `Collection`. This in particular enables the use of `Siblings.attach(_:on:edit:)` with pivots which use a composite ID (most often consisting of the two parent relations).

Additional changes:

- Swift 5.4 support removed.
- NIO deprecation warnings fixed.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.5
 import PackageDescription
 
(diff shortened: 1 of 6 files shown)


0xTim (approved):
Well that wasn't too bad!

PR #541, opened 2022-12-05
Fix return type to non optional of Databases.database()

sidepelican:
`Databases.database(...)` is specified to return `Database?` but it never returns `nil` so this PR fixes the signature.


gwynne:
I wish we could make this change, truly I do - this annoying inconsistency is the result of @tanner0101 not updating the API's return type when he changed the implementation all the way back in [Fluent 4 beta 5](https://github.com/vapor/fluent-kit/pull/170). Unfortunately, changing this would break public API 🙁.

PR #550, opened 2023-02-18
Additional relations support for CompositeID models

gwynne:
This update includes the following:

- `Model.joined(_:)` now works correctly with both aliased models and models with spaces.
- Lots more documentation comments in various places.
- `DatabaseInput` got a new flag to allow explicitly requesting that all properties - not just modified ones - be included. See the inline documentation for more details. All built-in FluentKit property types appropriately respect this flag.
- Fix completely incorrect implementation of `KeyPrefixingStrategy`'s `.camelCase` strategy, including adding unit tests.
- Removed outdated checks for Swift version and OS minimums in Concurrency support since we require those minimums already.
- `@Group` properties with unset values no longer crash when encoded.
- `@CompositeID` properties can now be encoded/decoded even when explicitly given `nil` value.
- Implemented `@CompositeOptionalParent`, `@CompositeChildren`, and `@CompositeOptionalChild` properties to go with `@CompositeParent`, all of which allow fully describing relations even when the parent side of the relationship has a composite ID.
- The minimum Swift version requirement has been bumped from 5.5 to 5.5.2. The recommended minimum version is 5.6.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,9 +60,9 @@ jobs:
             POSTGRES_HOST_AUTH_METHOD: scram-sha-256, POSTGRES_INITDB_ARGS: --auth-host=scram-sha-256
         }
       mongo-a:
-        image: mongo:5
+        image: mongo:6
       mongo-b:
-        image: mongo:5
+        image: mongo:6
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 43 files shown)


gwynne:
API breakage check remarks:
- Changes to the "public" types in FluentBenchmark are not true API breakage; only the top-level test methods should be invoked by dependents.
- The new protocol requirement in `DatabaseInput` has a public default implementation and does not require any changes by external code.

0xTim (changes requested):
🎉 only one thing that actually needs changing (the manifest) and a couple of questions

0xTim (approved):

PR #551, opened 2023-02-21
add filter to eager-loading.

laowudxf:
Add a filter param to `with` method of QueryBuilder+EeagerLoad to filter the eager-load result.
for example:
``` swift
    //sort person by name and nest eager load person‘s cats.
    // the tail closure is the nest eager load.
    let catesWithFilter = try Category.query(on: app.db).with(\.$persons, filter: {qb in
        _ = qb.sort(\.$name, .descending)
    }) { (qb) in
            qb.with(\.$cats)
        }.all()
        .wait()
```

### DIFF
--- a/Sources/FluentKit/Model/EagerLoad.swift
+++ b/Sources/FluentKit/Model/EagerLoad.swift
@@ -21,6 +21,12 @@ public protocol EagerLoadable {
         _ relationKey: KeyPath<From, Self>,
         to builder: Builder
     ) where Builder: EagerLoadBuilder, Builder.Model == From
+    
+    static func eagerLoad<Builder>(
+        _ relationKey: KeyPath<From, Self>,
+        filter: QueryBuilderFilterBlock<To>?,
+        to builder: Builder
+    ) where Builder: EagerLoadBuilder, Builder.Model == From
 
     static func eagerLoad<Loader, Builder>(
         _ loader: Loader,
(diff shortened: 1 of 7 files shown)


0xTim:
@laowudxf thanks for the PR! This will need lots of back and forth to land. First and foremost we need tests to cover all of the new behaviour. Second the PR only addresses primary key relations, we also need to account for composite key relations. Once those 2 issues are resolved we can go over the API, but at a high level I'm not overly comfortable with the syntax. As we already have the concept of closures for eager loading, it would be great to be able to reuse them and I'd love to be able to do something like:

```swift
let catesWithFilter = try await Category.query(on: app.db).with(\.$persons) {
  qb.with(\.$cats)
  qb.sort(\.$name, .descending)
}.all()

```

laowudxf:
@0xTim thanks for your reply, your considerations are deliberate. 
For the better syntax it's looks much better, but it would be take a big change to the origin code, I don't have the confidence to make such a big change.

madsodgaard:
Just adding my two cents. This would definitely be great to have in Fluent! But I don't think it is appropriate that we use `QueryBuilder` to add these custom filters. It has too many actions that don't make sense in the context of eager-loading. For example I can now use `.run()`, `.create(), `.all()` etc inside an eager-loading block. I feel like this needs a new type of "builder", that only really allows the operations that makes sense here.

laowudxf:
@madsodgaard , you are right! 
For your advice, I have an idea. 
create a  filter struct (maybe called EagerloadFilter) hoding buideQuery privately, and expose some valid filter method, then pass it to filter block, now user can only use these method that we exposed.

0xTim:
Yep I agree with @madsodgaard - we need to ensure that the queries built are valid. The `.group()` filter might be useful as inspiration - https://docs.vapor.codes/fluent/query/#group but that doesn't expose sorting etc

I want to loop in @gwynne as well as she owns the Fluent codebase and understands it far better that I do

gwynne:
This API is source-breaking in its current form, which is a non-starter regardless of other considerations.

PR #554, opened 2023-03-03
Enforce correct @ID(key:) usage at compile time

gwynne:
Using `@ID(key:)` with a key other than `.id` or a value that is not a `UUID` is now a compile-time error instead of a runtime crash.

Additional changes:

- The legacy `@ID(key: "id")` syntax, where the key is specified as a string rather than using the `.id` enumeration case, is hard-deprecated. Replace such usages with `@ID(key: .id)`. This deprecation does _not_ apply to custom IDs.
- The `@ID(key: .id)` syntax is now considered soft-deprecated. `@ID` is now the recommend usage.
- Minor CI improvements

## ⚠️ **IMPORTANT** ⚠️

1. Incorrect usage of `@ID(key: "...")` with any string other than `"id"` will still trigger a fatal error at runtime as well as the deprecation warning. This is unavoidable without a source-breaking change, since the contents of the string can not be checked at compile time.
2. Incorrect usage of `@ID` where the type of the property is not `UUID` - for example, `@ID var id: Int?` - usually results in the extremely unhelpful compiler error "`Type of expression is ambiguous without more context.`" Unfortunately, there doesn't seem to be a way to make this less opaque at the moment.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,7 @@
 name: test
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 on:
   pull_request: { branches: ['*'] }
   push: { branches: ['main'] }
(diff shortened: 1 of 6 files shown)


0xTim (approved):
Looks good!

Issue #543, opened 2022-12-28
Generated queries have non-deterministic ordering of columns, which trash monitoring/logging/performance


Remi-C:
**Is your feature request related to a problem? Please describe.**
We use this ORM to generate inserts in a postgres db.
Looking at the DB logs, the inserts have random order of columns.
The insert end up inserting the same dat, but is very bad for performance (makes it harder for the DB to cache plans),
and terrible for monitoring/analytics (the same query appears as dozens of different queries depending the order of columns)

here is an example of 10 different inserts actually doing the same thing
``` SQL
INSERT INTO my_table ("col1", "col2", "col3", "col4", "col5", "col6", "col7", "col8", "col9", "col10", "col11", "col12", "col13", "col14", "col15", "col16", "col17", "col18", "col19", "col20", "col21", "col22", "col23") VALUES ($1, $2, NULL, $3, $4, $5, $6, NULL, $7, DEFAULT, $8, $9, $10, $11, $12, $13, $14, NULL, $15, $16, $17, $18, $19) RETURNING "col10"
INSERT INTO my_table ("col9", "col7", "col16", "col19", "col2", "col18", "col15", "col13", "col3", "col5", "col1", "col10", "col22", "col6", "col8", "col21", "col12", "col23", "col11", "col20", "col4", "col14", "col17") VALUES ($1, $2, $3, $4, $5, NULL, $6, $7, NULL, $8, $9, DEFAULT, $10, $11, NULL, $12, $13, $14, $15, $16, $17, $18, $19) RETURNING "col10"
INSERT INTO my_table ("col21", "col2", "col18", "col10", "col12", "col20", "col9", "col8", "col15", "col6", "col5", "col11", "col13", "col22", "col7", "col16", "col14", "col3", "col4", "col23", "col17", "col19", "col1") VALUES ($1, $2, NULL, DEFAULT, $3, $4, $5, NULL, $6, $7, $8, $9, $10, $11, $12, $13, $14, NULL, $15, $16, $17, $18, $19) RETURNING "col10"
[...]

PR #556, opened 2023-03-09
make insert query column ordering deterministic

rausnitz:
When composing insert queries, the order of the columns is non-deterministic. This hurts query monitoring and analytics.

In a Postgres database we work with, `SELECT * FROM pg_stat_statements WHERE query like 'INSERT INTO "my_insert_heavy_table"%';` returns more than 3,100 rows. The relevant table `my_insert_heavy_table` has many columns, thus many possible column orders when it's non-deterministic. I think if we make the order deterministic, these 3,100 different queries could be treated as one query for monitoring purposes.

Resolves https://github.com/vapor/fluent-kit/issues/543


gwynne:
I had an idea for a better way to handle this which not only makes INSERT queries behave deterministically, but also actually improves the performance instead of worsening it. I implemented it on an identically-named branch which I will open a new PR for momentarily.

PR #555, opened 2023-03-08
Allow eager-loading of deleted models

tarag:
Allows eager-loading of soft-deleted models. For maximum flexibility, choosing whether to load soft-deleted models or not is performed for each relation at each query call.

Usage is as follows: `let galaxies = try Galaxy.query(on: self.database).with(\.$stars, withDeleted: true)`

Use of default argument value of false requires no change to existing code to keep existing behaviour.

Fixes #375.

PR #558, opened 2023-03-13
Added support for eager loading of soft deleted relations.

maciejtrybilo:
This change adds `withDeleted: Bool` parameter to the `with()` query building function. For example:

```
let galaxies = try Galaxy.query(on: self.database)
  .with(\.$stars, withDeleted: true)
  .all()
```

will also return stars that are soft deleted.

There are a few tests that I'd still add in EagerLoadTests, but I would like to check with the maintainers if such an approach would be welcome.

Technically it’s a breaking change as it adds the `withDeleted` property to the public `EagerLoader` protocol. In practice It seems to me few clients would be implementing it, so it shouldn't cause much grief if any.

### DIFF
--- a/Sources/FluentBenchmark/SolarSystem/Planet.swift
+++ b/Sources/FluentBenchmark/SolarSystem/Planet.swift
@@ -60,7 +60,7 @@ public struct PlanetSeed: Migration {
     public init() { }
 
     public func prepare(on database: Database) -> EventLoopFuture<Void> {
-        Star.query(on: database).all().flatMap { stars in
+        Star.query(on: database).withDeleted().all().flatMap { stars in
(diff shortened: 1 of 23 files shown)


maciejtrybilo:
I've been on it for a good while and just noticed that there is https://github.com/vapor/fluent-kit/pull/555 open already. Please feel free to close!

tarag:
It seems your proposal is more complete than mine since I stepped over the `Through...` loaders, and you have updated tests too.

If this is an issue, I have a non breaking-change where a new protocol is added for these `withDeleted` calls.

Anyway, I hope one of the other PR will move forward.

0xTim:
Closing in favour of #555

PR #567, opened 2023-06-09
Explicitly log migration errors

gwynne:
This ensures errors thrown during migration (both prepare and revert) are fully logged regardless of the behavior of calling code.


tkrajacic (approved):
LGTM

Andrewangeta (approved):
:shipit:

PR #568, opened 2023-06-10
added some missing `CustomStringConvertible` conformances

RandomHashTags:
Some properties already conform to this protocol, but I found some that do not. This fixes a case where getting the description of an `AnyProperty` will not be in the correct format, and also missing the key.

## Changes
- updated 5 properties: `OptionalBooleanProperty`, `BooleanProperty`, `OptionalFieldProperty`, `EnumProperty`, and `OptionalEnumProperty`
- updated 1 protocol (and its default implementations): `BooleanPropertyFormat`

## Examples
- an `@Enum` would return `FluentKit.EnumProperty<X, Y>`. Now returns `@X.Enum<Y>(key: Z)`
- an `@OptionalField` would return `FluentKit.OptionalFieldProperty<X, Y>`. Now returns `@X.OptionalField<Y>(key: Z)`

## Unit Testing
- This commit passed unit tests, except sometimes `FluentKitTests.testAnyModelDescriptionFormatHasNotChanged`, even though the data is the same, just in a different order (this test also fails even before this commit)

### DIFF
--- a/Sources/FluentKit/Properties/BooleanPropertyFormat.swift
+++ b/Sources/FluentKit/Properties/BooleanPropertyFormat.swift
@@ -1,5 +1,5 @@
 /// A conversion between `Bool` and an arbitrary alternative storage format, usually a string.
-public protocol BooleanPropertyFormat {
+public protocol BooleanPropertyFormat : CustomStringConvertible {
     associatedtype Value: Codable
 
     init()
@@ -23,6 +23,10 @@ public struct DefaultBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == DefaultBooleanPropertyFormat {
     public static var `default`: Self { .init() }
+    
+    public var description: String {
+        return "default"
+    }
 }
 
 /// Represent a `Bool` as any integer type. Any value other than `0` or `1` is considered invalid.
@@ -43,6 +47,10 @@ public struct IntegerBooleanPropertyFormat<T: FixedWidthInteger & Codable>: Bool
     
     public func serialize(_ bool: Bool) -> T {
         return .zero.advanced(by: bool ? 1 : 0)
+    }
+    
+    public var description: String {
+        return "integer"
     }
 }
 
@@ -69,6 +77,10 @@ public struct OneZeroBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == OneZeroBooleanPropertyFormat {
     public static var oneZero: Self { .init() }
+    
+    public var description: String {
+        return "oneZero"
+    }
 }
 
 /// Represent a `Bool` as the strings "N" and "Y". Parsing is case-insensitive. Serialization always stores uppercase.
@@ -90,6 +102,10 @@ public struct YNBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == YNBooleanPropertyFormat {
     public static var yn: Self { .init() }
+    
+    public var description: String {
+        return "yn"
+    }
 }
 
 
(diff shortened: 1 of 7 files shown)


Sources/FluentKit/Properties/BooleanPropertyFormat.swift
```diff
@@ -69,6 +77,10 @@ public struct OneZeroBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == OneZeroBooleanPropertyFormat {
     public static var oneZero: Self { .init() }
+    
+    public var description: String {
```
MahdiBM, line 81: `extension BooleanPropertyFormat where Self == OneZeroBooleanPropertyFormat` effectively resolves to `extension OneZeroBooleanPropertyFormat` so i think it's better to just use a normal extension like `extension OneZeroBooleanPropertyFormat: CustomStringConvertible` for the conformance just to make it possibly easier for the compiler, and also make it more straight-forward for other readers.  

The reason `extension BooleanPropertyFormat where Self == OneZeroBooleanPropertyFormat` is used at all, is for some compiler-required type inference in the `static var`s, otherwise it doesn't actually do much.
For example if you try:
```swift
extension BooleanPropertyFormat {
    public static var oneZero: Self { OneZeroBooleanPropertyFormat() }
}
```
it won't work (compiler will complain). but what is defined in the code does work because of that `where` clause.

Sources/FluentKit/Properties/BooleanPropertyFormat.swift
```diff
@@ -23,6 +23,10 @@ public struct DefaultBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == DefaultBooleanPropertyFormat {
     public static var `default`: Self { .init() }
+    
+    public var description: String {
```
MahdiBM, line 27: same as `OneZeroBooleanPropertyFormat`

Sources/FluentKit/Properties/BooleanPropertyFormat.swift
```diff
@@ -112,6 +128,10 @@ public struct YesNoBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == YesNoBooleanPropertyFormat {
     public static var yesNo: Self { .init() }
+    
+    public var description: String {
```
MahdiBM, line 132: same as `OneZeroBooleanPropertyFormat`

Sources/FluentKit/Properties/BooleanPropertyFormat.swift
```diff
@@ -133,6 +153,10 @@ public struct OnOffBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == OnOffBooleanPropertyFormat {
     public static var onOff: Self { .init() }
+    
+    public var description: String {
```
MahdiBM, line 157: same as `OneZeroBooleanPropertyFormat`

Sources/FluentKit/Properties/BooleanPropertyFormat.swift
```diff
@@ -154,6 +178,10 @@ public struct TrueFalseBooleanPropertyFormat: BooleanPropertyFormat {
 
 extension BooleanPropertyFormat where Self == TrueFalseBooleanPropertyFormat {
     public static var trueFalse: Self { .init() }
+    
+    public var description: String {
```
MahdiBM, line 182: same as `OneZeroBooleanPropertyFormat`

MahdiBM (changes requested):
This looks good to me. 

I saw you changed some more stuff like `value_type` -> `valueType` in other reverted commits ... we can't do those because `value_type` was part of a `public` protocol and if we do that, it will be a _breaking change_, meaning other people's code won't compile anymore since the property name has changed. Although i too think that property should have been named as `valueType`, we still need to wait for the next major version (fluent / fluent-kit 5) to be able to merge those changes

I suggest to keep this PR just for `CustomStringConvertible`  conformances, and if you have more suggestions for fluent-kit, do a new PR for those.

RandomHashTags:
Closing due to the `CustomStringConvertible` property conformance and associated implementation not really meaning to be there. A more accurate reasoning would be: I hold the same opinion shared by Gwynne during [discussion](https://discord.com/channels/431917998102675485/684159753189982218/1117378440870498334).

While I do support this pull request, I will not continue pursuing implementation due to it being a legacy design.

Issue #379, opened 2020-08-27
QueryBuilder.sum(KeyPath) always returns nil


Christian-Seiler:
I have a Class Course and try to use the `sum` method in the following way:

```swift
final class Course: Model, Content {
   static let schema: String  = "courses"
   ...
   @Field(key: "credits")
   var credits: Int
    ...
}


Course.query(on: database)
    .filter(...)
    .sum(\.$credits)
```

As outlined in the [documentation](https://docs.vapor.codes/4.0/fluent/query/#aggregate) I would expect the query to return type Int. 

> All aggregate methods except count return the field's value type as a result.

However, I receive an Optional<Int> instead. Furthermore, the result of it always is nil. When I query with the code below, I receive the correct value.

```swift
Course.query(on: database)
    .filter(...)
    .all()
    .map { (courses) -> Int in
          var total = 0
          courses.forEach { total += $0.credits }
          return total
    }
```

Craz1k0ek:
I tried reproducing your issue, but it appears to work fine for me.
The model and migration:
```swift
import Fluent
import Vapor

final class Course: Model, Content {
    static let schema: String  = "courses"
    
    @ID(key: .id)
    var id: UUID?
    
    @Field(key: "credits")
    var credits: Int
}

struct CourseMigration: Migration {
    func prepare(on database: Database) -> EventLoopFuture<Void> {
        database.schema(Course.schema)
            .id()
            .field("credits", .int)
            .create()
    }
    
    func revert(on database: Database) -> EventLoopFuture<Void> {
        database.schema(Course.schema).delete()
    }
}
```
I then proceeded to apply the migration in `configuration.swift`
```swift
app.databases.use(.sqlite(.file("db.sqlite")), as: .sqlite)
app.migrations.add(CourseMigration())
```
And for the `routes.swift`
```swift
func routes(_ app: Application) throws {
    app.get { req in
        return Course.query(on: req.db).filter(\.$credits < 100).sum(\.$credits) ?? -1
    }
    
    app.get("hello") { req -> EventLoopFuture<String> in
        let c = Course()
        c.credits = Int.random(in: 0 ..< 500)
        return c.save(on: req.db).transform(to: "OK")
    }
}
```
Are you sure you ran the migrations with `vapor run migrate` before issuing `vapor run`? What version of Vapor are you running? What database type do you use? If you copy and paste this code, does this problem still affect you?

Christian-Seiler:
[...]

PR #569, opened 2023-06-15
Handle aggregate types a very tiny bit better

gwynne:
This just adds explicit forwarding of the generic `Result` parameter of the two `QueryBuilder.aggregate()` variants which weren't doing so already. It was already being implicitly forwarded due to type inference, but better to do so explicitly.

Also adds notes to the `sum()` and `average()` sets of methods warning that the `Field.Value` return types are not always correct (usually only causes problems with Postgres) and giving the workaround. This is related to #379 and vapor/fluent-postgres-driver#166 but does NOT fix those issues; it only clarifies one of the ways to get around the problem.

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Aggregate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Aggregate.swift
@@ -25,48 +25,56 @@ extension QueryBuilder {
         self.aggregate(.count, key, as: Int.self)
     }
 
+    // TODO: `Field.Value` is not always the correct result type for `SUM()`, try `.aggregate(.sum, key, as: ...)` for now
     public func sum<Field>(_ key: KeyPath<Model, Field>) -> EventLoopFuture<Field.Value?>
         where Field: QueryableProperty, Field.Model == Model
     {
         self.aggregate(.sum, key)
     }
 
+    // TODO: `Field.Value` is not always the correct result type for `SUM()`, try `.aggregate(.sum, key, as: ...)` for now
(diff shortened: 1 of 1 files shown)


Sources/FluentKit/Query/Builder/QueryBuilder+Aggregate.swift
```diff
@@ -25,48 +25,56 @@ extension QueryBuilder {
         self.aggregate(.count, key, as: Int.self)
     }
 
+    // TODO: `Field.Value` is not always the correct result type for `SUM()`, try `.aggregate(.sum, key, as: ...)` for now
```
dannflor, line 28: Since these are public APIs, do you want to make these comments public DocC markup?

dannflor (changes requested):
Besides thinking the workaround should be broadcast publicly, LGTM

ptoffy (approved):

0xTim (approved):

Sources/FluentKit/Query/Builder/QueryBuilder+Aggregate.swift
gwynne: I thought about it, but if you want the honest fact of the matter, if I added it for these, I'd end up making myself do the _entire_ set of documentation for each method (params, description, etc.), including all the other methods that aren't affected by the problem. That's probably a good hour worth of work to do really properly (especially the way I write doc comments, just look around at some of other bits of Fluent I've documented 🤨), and I don't have it to spare right this moment 😰. Best I can do is say that if someone happens to add it to this branch for me, I will both gratefully and delightedly accept it 😅.

dannflor (approved):
Sorry for holding up the PR, probably should've just left a comment.

PR #572, opened 2023-07-11
Leverage the new support for nested path expressions in SQLKit

gwynne:
Fluent now takes generic advantage of the new support added by vapor/sql-kit#169 when generating nested subpath expressions.

(Unfortunately, Fluent has never actually invoked this code path to begin with, except via an explicit call to `.sql(json:_:)`, so it doesn't change much at the moment. However, this would always have been a prerequisite to make the real support for it in Fluent's API actually work, so it's a step in the right direction.)

`SQLJSONColumnPath` is now deprecated in favor of using SQLKit's new `SQLNestedSubpathExpression`.


MahdiBM (approved):
✅️

PR #574, opened 2023-07-21
Add possibility of nested eager load builder to throw

Frizlab:
This PR adds the possibility of throwing inside the block that adds nested eager loaders.
I need this because I add eager loading as needed depending on fields (user input), recursively, and I don’t know before it’s too late whether my fields are valid.

As the `with` functions that are modified are marked as `rethrows`, this should have no impact on existing code and should not require a major version bump.


gwynne (approved):
Not only does this not break source compatibility, the change is so purely additive I can justify marking it for a patch-level release 🙂 Thanks for contributing this!

PR #580, opened 2023-08-16
Correctly handle @OptionalEnum properties when using ~~ and != in filter()

### LINKED ISSUE #579: Filtering with ~~ on enum collection does not work
### Describe the bug
ORM fails to convert a valid filter to a valid SQL request for postgres when filtering on a field being contained in a collection of enum values.
This is akin to #385.

### To Reproduce
Here’s the minimal code (minus the decorum) to reproduce the issue, in a Postgres setup:
```
final class DbUser : Model {
   @ID
   var id: UUID?
   @Field(key: "sex")
   var sex: Sex
}
enum Sex : String, Hashable, Codable {
   case male
   case female
}
let sexes: Set<Sex> = [.male, .female]
try await DbUser.query(on: req.db).filter(\.$sex ~~ sexes).all()
```

### Expected behavior
The ORM should translate the request to SQL properly.
Instead, postgres complain with the error `No operator matches the given name and argument types. You might need to add explicit type casts., line: 656, message: operator does not exist: sex = text, position: 2389, routine: op_error`.

### Environment

* Vapor Framework version: 4.77.0
* Vapor Toolbox version: 18.7.1
* OS version: macOS 13.5
* Db: Postgres 14

### Additional context
N/A

gwynne:
We were incorrectly hardcoding the use of `.bind` query values when using the `~~` (contains) and `!~` (not contains) operators with properties having optional values in Fluent `filter()` expressions. This appears to be the only place where this was still happening.

Fixes #579.

### DIFF
--- a/Tests/FluentKitTests/AsyncTests/AsyncFilterQueryTests.swift
+++ b/Tests/FluentKitTests/AsyncTests/AsyncFilterQueryTests.swift
@@ -15,7 +15,7 @@ final class AsyncFilterQueryTests: XCTestCase {
         let db = DummyDatabaseForTestSQLSerializer()
         _ = try await Task.query(on: db).filter(\.$status == .done).all()
         XCTAssertEqual(db.sqlSerializers.count, 1)
-        XCTAssertEqual(db.sqlSerializers.first?.sql, #"SELECT "tasks"."id" AS "tasks_id", "tasks"."description" AS "tasks_description", "tasks"."status" AS "tasks_status" FROM "tasks" WHERE "tasks"."status" = 'done'"#)
+        XCTAssertEqual(db.sqlSerializers.first?.sql, #"SELECT "tasks"."id" AS "tasks_id", "tasks"."description" AS "tasks_description", "tasks"."status" AS "tasks_status", "tasks"."optional_status" AS "tasks_optional_status" FROM "tasks" WHERE "tasks"."status" = 'done'"#)
         db.reset()
     }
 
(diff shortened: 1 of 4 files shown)


Tests/FluentKitTests/AsyncTests/AsyncFilterQueryTests.swift
```diff
@@ -15,31 +15,64 @@ final class AsyncFilterQueryTests: XCTestCase {
         let db = DummyDatabaseForTestSQLSerializer()
```
MahdiBM, line 15: PR itself LGTM, but you should move these `let db = DummyDatabaseForTestSQLSerializer()` and `db.reset()` calls to `setup()` and `tearDown()` functions.

MahdiBM (changes requested):
Looks good other than a little bit of possible cleanup 🙂

Tests/FluentKitTests/AsyncTests/AsyncFilterQueryTests.swift
gwynne: That is nowhere remotely near important enough to be worth blocking a bugfix. If it was a new problem that'd be one thing, but the Fluent tests have been a mess and a half for a long time.

MahdiBM: Hmm i guess not blocking, but i'd like the change still 😝

MahdiBM (approved):
You're right the problem isn't blocking :)

PR #585, opened 2023-09-17
Allow pivot editing closures to be async and throwing.

### LINKED ISSUE #581: Async functions within attach closure
would be awesome if we could get async functionality within the attach closure as seen in this next example:
```swift
.. .attach(..., on: transaction, { example in
  let exampleID = example.id.value(or: Abort(.notFound))
  try await ExampleDetail(exampleID: exampleID).create(on: transaction)
}
```

atm this errors. but would be great if it were possible

gwynne:
When using the `attach(_:method:on:_:)` and `attach(_:on:_:)` methods of `SiblingsProperty`, the optional closure provided to allow editing new pivots before they are saved can now optionally be both `throws` and `async`.

Closes #581.

Additional changes:

- All versions of the `isAttached()`, `attach()`, and `detach()` methods of `SiblingsProperty` will now throw errors (specifically cases of the new `SiblingsPropertyError` enum) instead of calling `fatalError()` when an unsaved model is encountered.
- The `isAttached()` method now uses a much faster and more efficient query to check whether a pivot already exists for the given model.

### DIFF
--- a/Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
+++ b/Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
@@ -2,57 +2,104 @@ import NIOCore
 
 public extension SiblingsProperty {
     
-    func load(on database: Database) async throws {
+    func load(on database: any Database) async throws {
         try await self.load(on: database).get()
     }
     
     // MARK: Checking state
     
-    func isAttached(to: To, on database: Database) async throws -> Bool {
+    func isAttached(to: To, on database: any Database) async throws -> Bool {
         try await self.isAttached(to: to, on: database).get()
     }
     
-    func isAttached(toID: To.IDValue, on database: Database) async throws -> Bool {
+    func isAttached(toID: To.IDValue, on database: any Database) async throws -> Bool {
         try await self.isAttached(toID: toID, on: database).get()
     }
     
     // MARK: Operations
     
-    func attach(
-        _ tos: [To],
-        on database: Database,
-        _ edit: (Through) -> () = { _ in }
-    ) async throws {
+    /// Attach multiple models with plain edit closure.
+    func attach(_ tos: [To], on database: any Database, _ edit: (Through) -> () = { _ in }) async throws {
         try await self.attach(tos, on: database, edit).get()
     }
+
+    /// Attach single model with plain edit closure.
+    func attach(_ to: To, on database: any Database, _ edit: @escaping (Through) -> ()) async throws {
+        try await self.attach(to, method: .always, on: database, edit)
+    }
     
+    /// Attach single model by specific method with plain edit closure.
     func attach(
-        _ to: To,
-        method: AttachMethod,
-        on database: Database,
+        _ to: To, method: AttachMethod, on database: any Database,
         _ edit: @escaping (Through) -> () = { _ in }
     ) async throws {
         try await self.attach(to, method: method, on: database, edit).get()
     }
     
+    /// A version of ``attach(_:on:_:)-791gu`` whose edit closure is async and can throw.
+    ///
+    /// This method provides "all or none" semantics- if the edit closure throws an error, any already-
+    /// processed pivots are discarded. Only if all pivots are successfully edited are any of them saved.
+    ///
+    /// These semantics require us to reimplement, rather than calling through to, the ELF version.
     func attach(
-        _ to: To,
-        on database: Database,
-        _ edit: (Through) -> () = { _ in }
+        _ tos: [To],
+        on database: any Database,
+        _ edit: @Sendable @escaping (Through) async throws -> ()
     ) async throws {
-        try await self.attach(to, on: database, edit).get()
+        guard let fromID = self.idValue else {
+            throw SiblingsPropertyError.owningModelIdRequired(property: self.name)
+        }
+        
+        var pivots: [Through] = []
+        pivots.reserveCapacity(tos.count)
+        
+        for to in tos {
+            guard let toID = to.id else {
+                throw SiblingsPropertyError.operandModelIdRequired(property: self.name)
+            }
+            let pivot = Through()
+            pivot[keyPath: self.from].id = fromID
+            pivot[keyPath: self.to].id = toID
+            pivot[keyPath: self.to].value = to
+            try await edit(pivot)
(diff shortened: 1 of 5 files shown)


Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
```diff
@@ -2,57 +2,104 @@ import NIOCore
 
 public extension SiblingsProperty {
     
-    func load(on database: Database) async throws {
+    func load(on database: any Database) async throws {
```
MahdiBM, line 5: Perhaps we can also enable the flag in Package.swift ?

Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
```diff
@@ -2,57 +2,104 @@ import NIOCore
...
+    /// processed pivots are discarded. Only if all pivots are successfully edited are any of them saved.
+    ///
+    /// These semantics require us to reimplement, rather than calling through to, the ELF version.
     func attach(
-        _ to: To,
-        on database: Database,
-        _ edit: (Through) -> () = { _ in }
+        _ tos: [To],
+        on database: any Database,
+        _ edit: @Sendable @escaping (Through) async throws -> ()
```
MahdiBM, line 48: Since it's a new function, we don't even need the @preconcurrency 👌

Sources/FluentKit/FluentError.swift
```diff
@@ -1,13 +1,14 @@
 import Foundation
 
-public enum FluentError: Error, LocalizedError, CustomStringConvertible, CustomDebugStringConvertible {
+public enum FluentError: Error, LocalizedError, CustomNSError, CustomStringConvertible, CustomDebugStringConvertible {
```
MahdiBM, line 3: Why the `CustomNSError`? Is it for not getting those "Operation could not be completed" useless errors?

Sources/FluentKit/Properties/Siblings.swift
```diff
@@ -93,14 +93,14 @@ public final class SiblingsProperty<From, To, Through>
...
+            return database.eventLoop.makeFailedFuture(SiblingsPropertyError.owningModelIdRequired(property: self.name))
         }
 
         return Through.query(on: database)
             .filter(self.from.appending(path: \.$id) == fromID)
             .filter(self.to.appending(path: \.$id) == toID)
-            .first()
-            .map { $0 != nil }
+            .count()
+            .map { $0 > 0 }
```
MahdiBM, line 103: will be a nice performance boost 👌

Sources/FluentKit/Properties/Siblings.swift
```diff
@@ -117,20 +117,24 @@ public final class SiblingsProperty<From, To, Through>
...
     ) -> EventLoopFuture<Void> {
         guard let fromID = self.idValue else {
-            fatalError("Cannot attach siblings relation \(self.name) to unsaved model.")
+            return database.eventLoop.makeFailedFuture(SiblingsPropertyError.owningModelIdRequired(property: self.name))
         }
-
-        return tos.map { to -> Through in
+        
+        var pivots: [Through] = []
+        pivots.reserveCapacity(tos.count)
```
MahdiBM, line 124: I never went ahead to look into `map`'s implementation but i always assumed it must already do the `reserveCapacity` itself? Does it not do that?! If yes, do you know why?!

MahdiBM (changes requested):
The only reason for not approving is i want to know the answer to the `CustomNSError` comments first 🙂

Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
gwynne: That would require a secondary `Package@swift-5.8.swift` manifest and necessitate changes in almost every single file, I've been deliberately doing this gradually in most packages.

Sources/FluentKit/FluentError.swift
gwynne: Yes, it hopefully will improve error messages under some conditions on Linux; I can't prove conclusively that it does much of anything at all, but it isn't harmful or performance-affecting, so at worst it does nothing at all.

Sources/FluentKit/Properties/Siblings.swift
gwynne: It absolutely does. But that has nothing to do with why I didn't use `map()` here anyhow. Think about what the loop body has to deal with - this an old-style ELF implementation, remember; it can't just `throw` to terminate a `map()`.

Sources/FluentKit/FluentError.swift
MahdiBM: I don't think we need `CustomNSError` then if those exact useless errors i mentioned is what you're worrying about:
https://github.com/apple/swift/issues/58724#issuecomment-1693488876

Sources/FluentKit/Concurrency/Siblings+Concurrency.swift
MahdiBM: Fine

Sources/FluentKit/Properties/Siblings.swift
MahdiBM: 👌

Sources/FluentKit/FluentError.swift
gwynne: So just to check, you want me to remove a layer of fallback redundancy on the basis that you went ahead and "well actually"'d a Swift bug filed by _@weissi_ of all people on the assumption he couldn't possibly know that already? Do I have that right? (Please ensure your mental rendering of my tone in this comment expresses sufficiently over-the-top exaggerated snark for maximum comedic effect 😂)

MahdiBM: I wouldn't dare to do such thing against @weissi, that was more for others who might read the issue 😅

But yes, I do want you to remove that layer of fallback redundancy since i think it's pretty useless and just makes the code across different repos inconsistent (do you really want to, even gradually, add that `CustomNSError` to all other types in other repos as well, just for the slightest chance that it might do something, although you can try the thing i mentioned in the issue and see that's all you need? 😛)

MahdiBM (approved):
👌

PR #592, opened 2024-01-31
Add `@discardableResult` to QueryBuilder's range and sort methods

jiahan-wu:
This Pull Request addresses an issue where calling the sort methods in the QueryBuilder class would result in a "Result of call to 'sort' is unused" warning. To resolve this, I have added the @discardableResult attribute to these methods, allowing users to call them without triggering the warning when the result is not explicitly used.


gwynne (commented):
Thanks, nice catch!! The `range` methods in `QueryBuilder+Range.swift` also appear to be missing this attribute, could you add it to those as well?

PR #593, opened 2024-01-31
Relax restrictions on ModelAlias use and join conditions

### LINKED ISSUE #431: Filter joined ModelAlias(es) by shared properties

### LINKED ISSUE #536: Operator function '~~' doesn't accept a ModelAlias in QueryBuilder's filter
### Describe the bug

Subset filters are unusable when using ModelAliases. 

### To Reproduce

```swift
User.query(on: req.db)
    .join(UserProjectPivotAlias.self, on: \User.$id == \UserProjectPivotAlias.$user.$id)
    .filter(UserProjectPivotAlias.self, \UserProjectPivotAlias.$role ~~ [.owner, .editor, .member]) // <- Error
// Operator function '~~' requires that 'UserProjectPivotAlias' conform to 'Model'
```
Replacing the alias with the actual model works fine.
Below the other classes involved.
```swift
// Model Alias
final class UserProjectPivotAlias: ModelAlias {
    static let name = "user_project_alias"
    let model = UserProjectPivot()
}

// Model
final class UserProjectPivot: Model, Content {
    static let schema = "user_project"
    
    @ID(key: .id)
    var id: UUID?
    
    @Timestamp(key: "created_at", on: .create)
    var createdAt: Date?
    @Timestamp(key: "updated_at", on: .update)
    var updatedAt: Date?

    @Parent(key: "user_id")
    var user: User
    
    @Parent(key: "project_id")
    var project: Project
    
    @Field(key: "role")
    var role: Role

    init() { }

    init(id: UUID? = nil, user: User, project: Project, role: Role) throws {
        self.id = id
        self.$user.id = try user.requireID()
        self.$project.id = try project.requireID()
        self.role = role
    }
}
```

### Expected behavior

Models should be 100% replaceable with model aliases in query builders.

### Environment

* Vapor Framework version: 4.65.2
* Vapor Toolbox version: 18.5.1
* OS version: macOS 12.6

gwynne:
1. Relaxes most of Fluent's filter operators to only require `Schema` conformance rather than `Model`. This enables `ModelAlias` usage with all operators that it previously didn't work for.
2. Enables all model value filters in joins (e.g. `join(Foo.self, on: \Foo.$bar.$id == \Bar.$id && \Bar.$value > 4)`).
3. Adds `.sql(raw:)`, `.sql(embed:)`, and `.sql(_:)` for Fluent joins.

(Note: Not a semver-major break because the new generic requirements on operators are always strict supersets of the old ones.)

Supersedes #432. Fixes #431. Fixes #536.

### DIFF
--- a/Sources/FluentBenchmark/Tests/JoinTests.swift
+++ b/Sources/FluentBenchmark/Tests/JoinTests.swift
@@ -224,6 +224,12 @@ extension FluentBenchmarker {
                 .all().wait()
             
             XCTAssertFalse(planets.isEmpty)
+            
+            let morePlanets = try Planet.query(on: self.database)
+                .join(Star.self, on: \Planet.$star.$id == \Star.$id && \Star.$name != "Sun")
+                .all().wait()
+            
+            XCTAssertEqual(morePlanets.count, 1)
         }
     }
 }
--- a/Sources/FluentKit/Database/Database+Logging.swift
+++ b/Sources/FluentKit/Database/Database+Logging.swift
@@ -61,5 +61,5 @@ extension LoggingOverrideDatabase: SQLDatabase where D: SQLDatabase {
         self.database.execute(sql: query, onRow)
     }
     var dialect: SQLDialect { self.database.dialect }
-    var version: SQLDatabaseReportedVersion? { self.database.version }
+    var version: (any SQLDatabaseReportedVersion)? { self.database.version }
 }
--- a/Sources/FluentKit/Operators/FieldOperators.swift
+++ b/Sources/FluentKit/Operators/FieldOperators.swift
@@ -177,7 +177,7 @@ public func !=~ <Left, Right, LeftField, RightField>(
 }
 
 public struct ModelFieldFilter<Left, Right>
-    where Left: FluentKit.Model, Right: FluentKit.Model
+    where Left: FluentKit.Schema, Right: FluentKit.Schema
 {
     public init<LeftField, RightField>(
         _ lhs: KeyPath<Left, LeftField>,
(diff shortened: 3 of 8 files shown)


0xTim (approved):
Would be good to see a test to cover the new behaviour

0xTim (approved):

PR #600, opened 2024-04-19
conform FieldKey to Sendable

jaredh159:
There are perhaps other types that should be audited for `Sendable`-ity, but I'm not sure if there's a coordinated effort around that. I'm updating a Vapor app getting it ready for Swift 6 with all concurrency warnings turned to the max, and seeing warnings related to my usage of `FieldKey`, seems like an obvious candidate to be Sendable, although I could be missing some subtle nuance of Semver or something like that.

### DIFF
--- a/Sources/FluentKit/Properties/FieldKey.swift
+++ b/Sources/FluentKit/Properties/FieldKey.swift
@@ -34,3 +34,5 @@ extension FieldKey: CustomStringConvertible {
 extension FieldKey: Equatable { }
 
 extension FieldKey: Hashable { }
+
+extension FieldKey: Sendable { }


gwynne:
This will be superseded by #601

0xTim:
@jaredh159 I'm pretty sure most Vapor stuff should be ready outside of Fluent now. There are a couple in Vapor we can't fix, but most are done. Which are you seeing?

jaredh159:
@0xTim i might have spoke too soon on the vapor warnings, i wasn't at `4.93.0` yet when i wrote that comment, but have updated since and i think they're mostly (or all) gone, just the ones from fluent-kit remained.  if i see something from vapor core i'll let you know

PR #609, opened 2024-05-12
Tidy up FluentKit

gwynne:
A few changes that were missed in previous cleanup and update passes:

- Previously missing `async` versions of some `Database` methods are now present.
- All `SQLDatabase` methods are now correctly forwarded when using `Database.logging(to:)`.
- All versions of the `.sql(raw:)` utility for directly embedding raw SQL into Fluent queries have been renamed to `.sql(unsafeRaw:)` to better emphasize the inadvisability of their use. The old name still works but is now deprecated.
- `SQLSchemaConverter` is now very infinitesimally faster.
- Fluent's tests now include the appropriate tests from SQLKit as well.
- Removed some unneeded `Sendable` warning workarounds, should slightly improve performance.


Sources/FluentSQL/DatabaseSchema+SQL.swift
```diff
@@ -52,8 +67,13 @@ extension DatabaseSchema.ConstraintAlgorithm {
 }
 
 extension DatabaseSchema.FieldConstraint {
+    @available(*, deprecated, renamed: "sql(unsafeRaw:)", message: "Renamed to `.sql(unsafeRaw:)`. Please use caution when embedding raw SQL.")
```
0xTim, line 70: Feel like we should have a `SQLRawSupporting` protocol for all these so we don't need to duplicate everything

0xTim (approved):

Sources/FluentSQL/DatabaseSchema+SQL.swift
gwynne: Not worth the headache at this point, though yeah that would've been good design.

gwynne:
SQLite integration tests will pass once vapor/fluent-sqlite-driver#90 is merged

PR #615, opened 2024-08-06
joined

bwdmr:
joined try concession is misleading, expecting a throw if model does not exists. yet works flawlessly event model not in query. this separation between layout and joined keeps existing functionality and extends it by desired behaviour to disrupt execution once result does not meet expectation. layout is a plain copy of the function joined and its intend is to retrieve the layout of any model as a fluent object. ion the otherhand joined should retrieve the model from a query and throw once this not happens.


gwynne:
Unfortunately, this is a non-starter due to the break in API implied by adding additional protocol requirements to `.joined()`. Also, at this point development on Fluent 4 has basically stopped except for bug fixes, and this concern will not be relevant to Fluent 5.

PR #621, opened 2024-11-18
Upgrade Async Kit to 1.20.0

PWrzesinski:
Currently Fluent SQLite driver silently depends on features (specifically the `shutdownAsync` function) introduced in Async Kit 1.20.0. That dependency is transient, without specifying Async Kit in the SQLite driver's Package.swift, causing a compilation error when an earlier version (like 1.19.0) is used.

It would probably be more correct to specifically add that dependency in the SQLite driver's Package.swift, but personally I don't see anything wrong with just updating it here.


gwynne (approved):
Nice catch, thanks!

PR #622, opened 2024-12-26
feat(Migrations): added migration group protocol and method

WilliamFernsV3:
This is useful, because in big apps where there is hundreds of migrations, using `app.migrations.add` will make the `configure.swift` file really long. Instead, migration groups can be created for each model, and then simply adding the group to `app.migrations` using the `addGroup` method.

### DIFF
--- a/Sources/FluentKit/Migration/Migrations.swift
+++ b/Sources/FluentKit/Migration/Migrations.swift
@@ -1,8 +1,12 @@
 import NIOConcurrencyHelpers
 
+public protocol MigrationsGroup {
+    var migrations : [any Migration] { get set }
+} 
+
 public final class Migrations: Sendable {
     let storage: NIOLockedValueBox<[DatabaseID?: [any Migration]]>
-    
+
     public init() {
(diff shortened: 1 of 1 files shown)


gwynne:
While I do appreciate the utility of this addition, it's not something I think Fluent really needs to provide; anyone who needs it can add it in their own codebase as a trivial `extension` on `Migrations`. It's also worth noting that I've declared Fluent 4 essentially closed to new feature development in favor of work on Fluent 5 (though I freely admit I haven't done so in any really visible fashion, so it's questionable how I expect people to know that 😅).

WilliamFernsV3:
> While I do appreciate the utility of this addition, it's not something I think Fluent really needs to provide; anyone who needs it can add it in their own codebase as a trivial `extension` on `Migrations`. It's also worth noting that I've declared Fluent 4 essentially closed to new feature development in favor of work on Fluent 5 (though I freely admit I haven't done so in any really visible fashion, so it's questionable how I expect people to know that 😅).

Thanks.

PR #623, opened 2025-01-02
ensure the LEFT JOIN behaves as expected

bwdmr:
The WHERE clause will filter out these rows because NULL values do not satisfy either the IS NULL or the greater than conditions, effectively excluding them from the result set.

Turning the LEFT JOIN into an INNER JOIN for rows where NULL values would have been expected.

* move includeDeleted checks from query to join, where the filter will be applied
* copy filters over and add filter as necessary, on case deleted in not included
* keep includeDeleted checks within query aswell, as null checks now can exists in where clause aswell.
* adjust tests to reflect changes in code.


Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -208,6 +208,7 @@ extension Fields {
 }
 
 extension Schema {
+
```
gwynne, line 211: This and several other whitespace-only changes are unnecessary.

Sources/FluentKit/Properties/Timestamp.swift
```diff
@@ -222,4 +223,26 @@ extension Schema {
...
+        )
+        
+        var copy = filters
+        copy.append(.group([
+            .value(deletedAtField, .equal, .null),
+            .value(deletedAtField, .greaterThan, timestamp.currentTimestampInput)
+        ], .or))
+        
+        let filters = copy
+        return filters
```
gwynne, line 246:
```suggestion
        return copy
```

Sources/FluentKit/Query/Builder/QueryBuilder+Join.swift
```diff
@@ -36,7 +36,16 @@ extension QueryBuilder {
...
+      
+        // If deleted models aren't included, add filters
+        // to exclude them for each model being queried.
+        if !self.includeDeleted {
+            let filters = foreign.excludeDeleted(from: filters)
+            
+            return self.join(Foreign.self, on: .advancedJoin(schema: Foreign.schema, space: Foreign.space, alias: Foreign.alias, method, filters: filters))
+        }
+        
+        return self.join(Foreign.self, on: .advancedJoin(schema: Foreign.schema, space: Foreign.space, alias: Foreign.alias, method, filters: filters))
```
gwynne, line 48:
```suggestion
        let finalFilters: [DatabaseQuery.Filter]
        
        // If deleted models aren't included, add filters to exclude them for each model being queried.
        if !self.includeDeleted {
            finalFilters = foreign.excludeDeleted(from: filters)
        } else {
            finalFilters = filters
        }

        return self.join(Foreign.self, on: .advancedJoin(schema: Foreign.schema, space: Foreign.space, alias: Foreign.alias, method, filters: finalFilters))
```

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -303,10 +303,11 @@ public final class QueryBuilder<Model>
                 self.addFields(for: model, to: &query)
             }
         }
-
+        
+        
         // If deleted models aren't included, add filters
         // to exclude them for each model being queried.
-        if !self.includeDeleted {
+        if  !self.includeDeleted {
```
gwynne, line 310: Unneeded whitespace change

Sources/FluentKit/Query/Database/DatabaseQuery+Action.swift
```diff
@@ -1,5 +1,24 @@
...
+                guard type(of: lhs) == type(of: rhs) else { return false }
+                return String(describing: lhs) == String(describing: rhs)
+            case let (.custom(lhs), .custom(rhs)):
+                guard type(of: lhs) == type(of: rhs) else { return false }
+                return String(describing: lhs) == String(describing: rhs)
+            default:
+                return false
+            }
+        }
+
```
gwynne, line 21: Why is this here? It does not appear to be needed by this PR, nor is it correct even if it were.

gwynne (changes requested):

Sources/FluentKit/Properties/Timestamp.swift
bwdmr: I recommend implementing an additional validation step for memory safety. To ensure proper memory allocation and prevent unsafe variable values, we should first convert the input into an immutable constant using the 'let' declaration. This approach provides better control over data handling and reduces potential security risks.

0xTim: That's not how Swift works, there's no issue around memory safety and since it gets returned anyway the compiler will optimise the lines out. There's no improvement over data handling or security and frankly just looks odd to read

bwdmr: i see my mistake,

When a variable is returned from a function, its mutability (var vs let) does not carry over. The mutability of the returned value depends on how it's captured in the receiving context.

The mutability of the original variable inside the function does not affect how the returned value can be used outside the function.

gwynne:
It appears that the Mongo driver (or Mongo itself) can't handle this. @Joannis, can you provide any insight?

Sources/FluentKit/Query/Builder/QueryBuilder.swift
```diff
@@ -303,7 +303,7 @@ public final class QueryBuilder<Model>
                 self.addFields(for: model, to: &query)
             }
         }
-
+
```
gwynne, line 306: Excess whitespace change, please remove

bwdmr:
support for complex joins with filters is not implemented yet.

https://github.com/vapor/fluent-mongo-driver/blob/79bf35430e72e4290b3d08ac1eb8784d4e389c32/Sources/FluentMongoDriver/MongoDB%2BJoin.swift#L114

joins with filter would require a pipeline as in the following example
```swift
db.orders.aggregate([
  { $lookup: {
    from: "customers",
    let: { 
      orderId: "$customerId",
      orderDate: "$orderDate" 
    },
    pipeline: [
      { $match: {
        $expr: {
          $and: [
            { $eq: ["$_id", "$$orderId"] },
            { $gt: ["$memberSince", "$$orderDate"] },
            { $eq: ["$isVerified", true] }
          ]
        }
      }}
    ],
    as: "customerInfo"
  }}
])
```

gwynne:
A solution for the Mongo driver is being worked on (though not by me, as I have no knowledge of Mongo 😅)

bwdmr:
> A solution for the Mongo driver is being worked on (though not by me, as I neither have nor want any knowledge of Mongo 😅)

can this be merged. i believe the mongo driver to be deprecated

gwynne:
Although the Mongo driver is deprecated for use in new Fluent 4 projects, it is unfortunately still in use by existing projects, and merging this would break normal usage for those users.

bwdmr:
can we proceed. support for the mongo community could be included in a separate pr

PR #628, opened 2025-02-06
fix: handle nil CompositeIDProperty value during database insert

bwdmr:
The CompositeIDProperty was force-unwrapping its value during database input, which would crash if the value was nil. This could happen during initial inserts where the composite ID is meant to be set by a database trigger. Now checks if value exists before attempting to input it.

### DIFF
--- a/Sources/FluentKit/Properties/CompositeID.swift
+++ b/Sources/FluentKit/Properties/CompositeID.swift
@@ -51,7 +51,9 @@ extension CompositeIDProperty: AnyDatabaseProperty {
     }
 
     public func input(to input: any DatabaseInput) {
-        self.value!.input(to: input)
+        if let value = self.value {
+            value.input(to: input)
+        }
     }
 
     public func output(from output: any DatabaseOutput) throws {


gwynne:
How are you encountering this scenario? The composite ID value is default-initialized to avoid exactly this failure mode.

bwdmr:
i have a trigger which creates the id on demand, given the composite id's members consist of
id: Int, sequenceNo: Int, timestamp: Int. i do need a noninitialized composite id so the db can create it.

bwdmr:
idkwym, i have a db function and a trigger on insert.
i need the composite id to be able to be nil on init so i can make use of the trigger.

gwynne:
The auto-initialized composite ID contains no values, so no values are (or can be) generated or sent by Fluent. In short, its presence should not prevent your trigger from filling them in itself. If it is, something else is wrong.

gwynne (approved):
Upon further examination, this is actually a valid fix, and avoids a crash in cases where the id of a model with composite ID has been explicitly set to `nil` (which was not previously meaningful, but with this change now becomes equivalent to setting it to `.init()`). Thanks for the fix and for getting me to take another look!

PR #633, opened 2025-03-29
Add the missing `.sql()` helpers for various DatabaseQuery types to FluentSQL

gwynne:
Adds `.sql(unsafeRaw:)`, `.sql(embed:)`, and `.sql(_:)` helpers for the following types to FluentSQL:
- `DatabaseQuery.Action`
- `DatabaseQuery.Aggregate`
- `DatabaseQuery.Aggregate.Method`
- `DatabaseQuery.Filter.Method`
- `DatabaseQuery.Filter.Relation`
- `DatabaseQuery.Join.Method`
- `DatabaseQuery.Sort.Direction`

Additional changes:
- `Schema.spaceIfNotAliased` is now `public` (it was previously `internal`).
- Updated the minimum version requirements for a couple of dependencies and made the `FluentKit` target's product dependencies more explicit.
- `DatabaseQuery.Limit.custom(_:)` and `DatabaseQuery.Offset.custom(_:)` are now deprecated, reflecting the fact that any use of them causes an unconditional `fatalError()`. Due to design limitations in SQLKit, it is not possible to implement them correctly instead of erroring.
- As promised, the use of `@_spi` has been removed in favor of the `package` access modifier now that we require Swift 5.9. This removes public symbols from the API, but because those symbols were SPI and had been explicitly documented as becoming non-public as of 5.9, this does not constitute a `semver-major` break. Anyone who was using the SPIs did so at their own risk.
- Fixed a missed `#if` condition for one of the bits of `Mirror` bypass logic (it still said `<6` instead of `<6.1`).
- A focused code formatting pass was made to remove lots of redundant uses of the `return` keyword, pretty much for the sole reason that I felt like doing it. Fixed a few incorrect comments and code style quirks along the way too, but only a few.

### DIFF
--- a/Sources/FluentKit/Model/Schema.swift
+++ b/Sources/FluentKit/Model/Schema.swift
@@ -11,7 +11,7 @@ extension Schema {
         self.alias ?? self.schema
     }
     
-    static var spaceIfNotAliased: String? {
-        return self.alias == nil ? self.space : nil
+    public static var spaceIfNotAliased: String? {
+        self.alias == nil ? self.space : nil
     }
 }
--- a/Package.swift
+++ b/Package.swift
@@ -16,8 +16,8 @@ let package = Package(
         .library(name: "XCTFluent", targets: ["XCTFluent"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.79.0"),
-        .package(url: "https://github.com/apple/swift-log.git", from: "1.6.2"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.81.0"),
+        .package(url: "https://github.com/apple/swift-log.git", from: "1.6.3"),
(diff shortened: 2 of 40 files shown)


Sources/FluentKit/Model/Schema.swift
```diff
@@ -11,7 +11,7 @@ extension Schema {
         self.alias ?? self.schema
     }
     
-    static var spaceIfNotAliased: String? {
-        return self.alias == nil ? self.space : nil
+    public static var spaceIfNotAliased: String? {
```
ptoffy, line 14: Is this intended?

ptoffy (approved):

Sources/FluentKit/Model/Schema.swift
gwynne: It certainly is. But I completely forget to mention it in the commit log or the PR description 😅

PR #637, opened 2025-05-05
Fix arithmetic  overflow crash in request pagination

petrpavlik:
When paging request, providing large values for `page` and `per` params will cause an arithmetic overflow exception and crash the process

Fixes a crash where trying to paginate with values that when multiplied cause an arithmetic overflow exception and crash the process. Following added unit tests provides an example.
```swift
    func testPaginationDoesntCrashOnOverflow() throws {
        let pageRequest1 = PageRequest(page: 1184467440737095516, per: 1184467440737095516)
        db.fakedRows.append([.init(["aggregate": 1])])
        // This would crash on `Swift runtime failure: arithmetic overflow` if not handled
        // so no point trying any XCTAssert
        _ = try Planet2
            .query(on: db)
            .paginate(pageRequest1)
            .wait()
    }
```

### DIFF
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -136,10 +136,10 @@ public struct PageRequest: Decodable, Sendable {
     }
 
     var start: Int {
-        (self.page - 1) * self.per
+        (self.page - 1).multipliedReportingOverflow(by: self.per).partialValue
     }
 
     var end: Int {
-        self.page * self.per
+        self.page.multipliedReportingOverflow(by: self.per).partialValue
     }
 }
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -140,6 +140,6 @@ public struct PageRequest: Decodable, Sendable {
     }
 
     var end: Int {
-        self.page.multipliedReportingOverflow(by: self.per).partialValue
+        self.page &* self.per
     }
 }
--- a/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
+++ b/Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
@@ -136,7 +136,7 @@ public struct PageRequest: Decodable, Sendable {
     }
 
     var start: Int {
-        (self.page - 1).multipliedReportingOverflow(by: self.per).partialValue
+        (self.page - 1) &* self.per
     }
 
     var end: Int {
--- a/Tests/FluentKitTests/FluentKitTests.swift
+++ b/Tests/FluentKitTests/FluentKitTests.swift
@@ -717,16 +717,16 @@ final class FluentKitTests: XCTestCase {
             .wait())
     }
     
-    func testPaginationDoesntCrashOnOverflow() throws {
+    func testPaginationDoesntCrashOnOverflow() async throws {
         let db = DummyDatabaseForTestSQLSerializer()
         let pageRequest1 = PageRequest(page: 1184467440737095516, per: 1184467440737095516)
         db.fakedRows.append([.init(["aggregate": 1])])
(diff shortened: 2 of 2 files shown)


petrpavlik:
> How in the world did you run into this in the first place? 😳

Tried after reading this https://forums.swift.org/t/what-are-the-pain-points-of-deploying-swift-on-server-applications/76435/15

0xTim:
We should be able to catch these things with some automated fuzzing at some point in the future

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
```diff
@@ -136,10 +136,10 @@ public struct PageRequest: Decodable, Sendable {
     }
 
     var start: Int {
-        (self.page - 1) * self.per
+        (self.page - 1).multipliedReportingOverflow(by: self.per).partialValue
```
gwynne, line 139: If we're just going to use the partial value result anyway, don't bother going through `multipliedReportingOverflow(by:)`, just disable the runtime check by requesting modulus arithmetic:
```suggestion
        (self.page - 1) &* self.per
```

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
```diff
@@ -136,10 +136,10 @@ public struct PageRequest: Decodable, Sendable {
...
     }
 
     var end: Int {
-        self.page * self.per
+        self.page.multipliedReportingOverflow(by: self.per).partialValue
```
gwynne, line 143: Same as above
```suggestion
        self.page &* self.per
```

Tests/FluentKitTests/FluentKitTests.swift
```diff
@@ -717,6 +717,18 @@ final class FluentKitTests: XCTestCase {
...
+        let pageRequest1 = PageRequest(page: 1184467440737095516, per: 1184467440737095516)
+        db.fakedRows.append([.init(["aggregate": 1])])
+        // This would crash on `Swift runtime failure: arithmetic overflow` if not handled
+        // so no point trying any XCTAssert
+        _ = try Planet2
+            .query(on: db)
+            .paginate(pageRequest1)
+            .wait()
+    }
+
```
gwynne, line 731: It's true it will crash if it fails, but we can still assert that we get a predictable result from it:
```suggestion
    func testPaginationDoesntCrashOnOverflow() async throws {
        let db = DummyDatabaseForTestSQLSerializer()
        let pageRequest1 = PageRequest(page: 1184467440737095516, per: 1184467440737095516)
        db.fakedRows.append([.init(["aggregate": 1])])
        let result = try await Planet2
            .query(on: db)
            .paginate(pageRequest1)
        XCTAssertEqual(result.metadata.page, 1184467440737095516)
        XCTAssertEqual(result.metadata.per, 1184467440737095516)
        XCTAssertEqual(result.metadata.total, 1)
    }
    
```

gwynne (changes requested):
Good fix, just some nitpicks on the details 🙂

gwynne:
Come to think of it, I wonder if we should be going for saturation rather than modulus overflow for this - if the user requests such an outlandish out-of-bounds value and we do overflow, the results effectively loop repeatedly; if we saturate, they stop getting results. The latter feels like a more correct response to this scenario to me.

petrpavlik:
What I'm trying to fix is an attacker being able to disable my backend by setting up a cron job that periodically performs a get request against any of my endpoints that support pagination with evil values.

I don't have any strong opinions on what values should be computed in such case. As you say, such values will never be used for anything legit anyway.

Sources/FluentKit/Query/Builder/QueryBuilder+Paginate.swift
petrpavlik: Nice, didn't know this existed to be honest.

Tests/FluentKitTests/FluentKitTests.swift
petrpavlik: Sure, happy to do that

PR #652, opened 2026-03-27
Enable `Foo.query(on:).set(\.$foo.$bar, to: baz).update()` with `@Group` properties

gwynne:
**These changes are now available in [1.56.0](https://github.com/vapor/fluent-kit/releases/tag/1.56.0)**

It is now possible to use the `.set(_:to:)` modifier on a query builder when referencing a property contained within an `@Group` property, e.g.:

```swift
final class FlatMoon: Model, @unchecked Sendable {
    final class Planet: Fields, @unchecked Sendable {
        enum PlanetType: String, Codable { case smallRocky, gasGiant, dwarf }

        @Field(key: "name") var name: String
        @Field(key: "type") var type: PlanetType
    }

    static let schema = "moons"

    @ID(key: .id) var id: String?
    @Group(key: "planet") var planet: Planet

    init() {}
}

try await FlatMoon.query(on: self.database)
    .set(\.$planet.$type, to: .dwarf) // <-- this works now!
    .filter(\.$id == "Moon")
    .update()
```
Also enables the Mirror bypass for 6.3 because why not. 🙂

Thanks to @mattesmohr for reporting the problem!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,6 @@ on:
   push: { branches: [ main ] }
 
 env:
-  LOG_LEVEL: info
   POSTGRES_HOSTNAME: 'psql-a'
   POSTGRES_HOSTNAME_A: 'psql-a'
   POSTGRES_HOSTNAME_B: 'psql-b'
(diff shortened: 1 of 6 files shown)


MahdiBM (approved):
LGTM

0xTim (approved):

